PR #1, opened 2018-08-07
Fix removing gradient steps from colorize node

TeddyDD:
Double click adds step, right click removes step. It seems that double click doesn't work with RMB.
This PR makes node a little bit bigger to contain whole gradient editor.

screen-1533638283


RodZill4:
TeddyDD, could you tell me which platform you are using ? RMB double click works for me under Windows 10.

TeddyDD:
Ubuntu and Void Linux. It might be a Godot bug, I'll try to pinpoint it and report on Godot repo. Thanks :)

PR #11, opened 2019-10-16
Add button to turn on rotation animation in preview window

jackaperkins:
Hi, please excuse my code, i'm very new to godot but I thought this would be a very simple addition, just a button that lets you toggle the rotation animation back on again, i found it frustrating that there was no way to do this.  If you like this idea but would prefer the UI works differently, let me know, i'm happy to change things.

rotate


jackaperkins:
Also I should say, I only tested this in godot editor, I dont know how you're making the standalone build, but i would hope this is so simple there's not much that could go wrong between the two formats.

PR #23, opened 2019-10-17
Fix recently added mixed indent in `main_window.gd`

Calinou:
It seems spaces were accidentally added (they weren't in the original PRs though).


RodZill4:
Hmm time to switch to Godot 3.2a2...

PR #26, opened 2019-10-18
Improve filtering in the library

Calinou:
The filter box is now located at the top of the library tree, as most search boxes are placed above the elements they search in.

Pressing Ctrl + F will now focus the filter box as well.

### DIFF
--- a/addons/material_maker/library.gd
+++ b/addons/material_maker/library.gd
@@ -4,6 +4,7 @@ extends VBoxContainer
 var libraries = []
 
 onready var tree : Tree = $Tree
+onready var filter_line_edit : LineEdit = $HBoxContainer/Filter
 
 func _ready():
 	tree.set_column_expand(0, true)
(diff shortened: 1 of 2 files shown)


RodZill4:
Hmm Ctrl+F is already used for making nodes editable (not a very good idea in fact. Maybe Ctrl+E would be a better choice).
Maybe we could change the filter-focus to spacebar (which is used for a similar action in Blender, substance designer...) ?

RodZill4:
Maybe also focus and select the contents (instead of just focus). So the contents can be deleted by just typing or kept by just pressing the right arrow key.

Calinou:
> Hmm Ctrl+F is already used for making nodes editable (not a very good idea in fact. Maybe Ctrl+E would be a better choice).

Indeed, I think moving the editable shortcut to <kbd>Ctrl + E</kbd> is a better idea.

> Maybe also focus and select the contents (instead of just focus). So the contents can be deleted by just typing or kept by just pressing the right arrow key.

I forgot about doing that, sorry. See #28.

PR #33, opened 2019-10-19
Enable 16× anisotropic filtering in the 3D preview

Calinou:
This improves rendering quality, especially when viewing the texture at an oblique angle.

### Before

Without anisotropic filtering

### After

With anisotropic filtering


RodZill4:
Hey ! It seems you started creating new sample materials. Please don't hesitate to add them to the example directory (with a comment to credit yourself in the graph).

PR #42, opened 2019-10-20
Increase the texture size to 2048×2048 for all examples

Calinou:
This makes example textures look better.

This also fixes the example texture sizes being too low after #38.

**Don't merge yet – see the comment below.**


Calinou:
I realized we may want to store the `size` attribute as an absolute power of two value, rather than an OptionButton index. This way, it won't break if we change the OptionButton's minimum value.

I can open a PR for this, and amend this one if you like the idea :slightly_smiling_face:

RodZill4:
> I realized we may want to store the `size` attribute as an absolute power of two value, rather than an OptionButton index.

Yes, this is how it should be, I guess.

RodZill4:
I just pushed the size parameter changes (does not depend on the minimum value of the parameter anymore) and resized all examples to 2048x2048.

Calinou:
Thanks!

I guess this PR can be closed then :slightly_smiling_face:

PR #47, opened 2019-10-27
Pretty-print the .ptex JSONs for better diffs in version control

Calinou:
This makes them about 15% larger on average, but we get better VCS diffs in exchange.

This doesn't break compatibility with existing materials.

All examples were resaved with this change. This also fixes the normal map filter texture sizes in all examples.


RodZill4:
This is a very good idea. I'm pretty sure .mmg and library files would also benefit a lot from being pretty printed...
The problem with normal maps was caused by a change in normal_map.mmg (old parameters changed names, so this broke compatibility). I'm afraid merging this as-is will break the samples again.
I can change this, you already spent time updating those examples...

Calinou:
@RodZill4 Sounds good to me. Feel free to apply the pretty-printing changes to MMG and other files too :slightly_smiling_face:

RodZill4:
I'm modifying this ATM.

RodZill4:
Done and pushed.

PR #49, opened 2019-10-27
Add SVG as a loadable image format

Calinou:
This makes use of Godot's SVG importer to load SVGs and rasterize them.


RodZill4:
Damn I didn't realize this worked already ! 👍

PR #56, opened 2019-11-20
Improve README with better formatting and license information

Calinou:
To be merged with #55.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,23 @@
-This is a tool based on the Godot game engine that can be used to create textures procedurally. It can also be used as a Godot addon (but there's no good reason for not using it as a standalone tool).
+# Material Maker
 
-Its user interface is based on Godot's GraphEditor: textures are described as interconnected texture generators and operators.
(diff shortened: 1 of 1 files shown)


RodZill4:
@Calinou There's a good reason for using it as a plugin: it will import .ptex files automatically, and you can use them as if they were SpatialMaterials (and they will be rendered at runtime).
there's also an option to prerender them, but there's a problem with how the plugin interacts
with the resource_filesystem (not sure what to do with this).

Plus I like to have as many tools as possible in Godot (did you have a look at my godot-material-spray project ?).

Calinou:
> There's a good reason for using it as a plugin: it will import .ptex files automatically, and you can use them as if they were SpatialMaterials (and they will be rendered at runtime).

I just kept the existing notice from the current README, I didn't check whether it was up-to-date :slightly_smiling_face:

RodZill4:
Just wanted to make sure you know, so you can test it and tell me how this could be improved. ;)

PR #57, opened 2019-11-20
Add `.desktop` and AppStream metadata files for Linux integration

Calinou:
These could be pulled into a Flatpak in the future.


RodZill4:
@calinou How does that affect Linux release delivery ? Should I package in a different manner ?

Calinou:
@RodZill4 The current way of doing is fine, this is mostly useful for those who intend to package Material Maker for Linux distributions or [Flathub](https://flathub.org/).

PR #74, opened 2020-01-29
Added popup menu for adding nodes on graph edit with space key

metincetin:


RodZill4:
I was planning to add this feature, so thanks a lot!

There are a few things I'd do differently, so I'd like your opinion:
- use a treeview (with everything expanded and no collapsing allowed) instead of a list
- maybe smaller icons so we can see more options
- create the node with a single click (and not a double-click)
- pop the window so it's completely visible when invoked at the bottom of the window
- create the node where the mouse pointer was when showing the popup and not when the node type is selected
- show the popup when connecting an output to empty space: in this case, the new node is created *and* connected

Please tell me what you think.

RodZill4:
Oh, and maybe not create a menu for each graph view but only one for the application.

metincetin:
Yes, your ideas makes sense alot. Especially liked that showing popup on empty space. Will be working on these.

RodZill4:
@metincetin Hmm not sure the treeview is a good idea, in fact. Maybe this list with full names (x/y/z) would be better. And support for multiple words in the search field (so you could search for "3d translate" for example).

metincetin:
You mean, finding the results that contain any word in filter like "3d" or "translate"? Or just finding exact match as "3d translate" (which it is doing now)

RodZill4:
I mean find nodes that contain both 3d and translate. 
I implemented that small change, with single click node creation, smaller icons and long node names: [add_node_popup.zip](https://github.com/RodZill4/material-maker/files/4136362/add_node_popup.zip)

metincetin:
Check it now, I implemented your changes, and the other features mentioned.

RodZill4:
Excellent! Do you have a twitter account that I can mention when communicating about this great new feature?

metincetin:
its @metnc_

PR #91, opened 2020-03-05
Fix screenshot path in README file

YeldhamDev:


RodZill4:
@YeldhamDev Thanx, just didn't notice this when I moved all files around. :D

PR #102, opened 2020-03-30
New Icon/Logo and Loading Screen

Hopefullyidontgetbanned:
Hi, I like this project so i designed a new Icon/Logo for it and tried to implement it natively, i'm new to godot so excuse me for any errors i made

The Splash Screen is identical however i changed the loading screen to this:
Screenshot (1329)
and added a new icon for project browser and for compiled binaries
icon


RodZill4:
I intended to replace the RodzLabs logo with something else in the loading screen, but not necessarily a logo (see Gimp, Blender... splash screens).
Not sure about the logo itself either. Any other idea ?

Hopefullyidontgetbanned:
it depends i Could change it to have a mesh with a procedral material made in material maker with the  Logo and text next to it however that would take longer.

RodZill4:
As an extra challenge, let's make it entirely in Material Maker. ;)

For example:
logo_albedo

Jummit:
I'd like it clean. Something like this would look good imo:
bitmap

Hopefullyidontgetbanned:
I could do something like that

Calinou:
I'd also prefer a flat/mostly flat shape as it can be represented with a SVG if we need it. With a screenshot of a procedural material, we can't regenerate it in an higher resolution easily (and it might not scale down well).

RodZill4:
Although regenerating materials in different resolutions with Material Maker is a piece of cake, I like @Jummit simple proposal. Maybe stick a Godot icon somewhere?

Jummit:
Something like this?
start_screen

RodZill4:
@Jummit Yes. Maybe the other logo ?
https://github.com/godotengine/godot/blob/master/editor/icons/Logo.svg

RodZill4:
Just modified the loading screen to:
loading

PR #107, opened 2020-04-21
Fix syntax error

### LINKED ISSUE #106: Cant Enable plugin due to Errors
**godot version**
3.2.2 beta 1

**OS version**
ubuntu 19.10 64x

**Problem**
Cant enable the plugin due to errors

Screenshot from 2020-04-20 09-58-37

tried to change to load(), but doesnt work either. There are some errors about mm_loader inside the import_plugin folder too

Relintai:
Hi!

This fixes the error "Parse Error: The argument name "__" is defined multiple times." on the latest engine.

closes #106 (this was the root cause)

### DIFF
--- a/addons/material_maker/engine/gen_base.gd
+++ b/addons/material_maker/engine/gen_base.gd
@@ -264,7 +264,7 @@ func get_shader_code(uv : String, output_index : int, context : MMGenContext) ->
 		print(rv)
 	return rv
 
-func _get_shader_code(__, __, __) -> Dictionary:
+func _get_shader_code(_uv, _output_index, _context) -> Dictionary:
 	return {}
 
 


addons/material_maker/engine/gen_base.gd
```diff
@@ -264,7 +264,7 @@ func get_shader_code(uv : String, output_index : int, context : MMGenContext) ->
 		print(rv)
 	return rv
 
-func _get_shader_code(__, __, __) -> Dictionary:
+func _get_shader_code(uv, output_index, context) -> Dictionary:
```
Calinou, line 267: Prefix unused parameters with an underscore to avoid GDScript warnings:

```suggestion
func _get_shader_code(_uv, _output_index, _context) -> Dictionary:
```

Relintai: ah, thanks

PR #121, opened 2020-05-23
Remove != null and == null checks

Jummit:
This is just a suggestion, and may break things.
Before:
```gdscript
if path == null:
	return
if generator != null:
	generator.set_parameter("image", path)
```
After:
```gdscript
if not path:
	return
if generator:
	generator.set_parameter("image", path)
```

### DIFF
--- a/addons/material_maker/engine/gen_base.gd
+++ b/addons/material_maker/engine/gen_base.gd
@@ -182,7 +182,7 @@ func get_source(input_index : int) -> OutputPort:
 
 func get_targets(output_index : int) -> Array:
 	var parent = get_parent()
-	if parent != null:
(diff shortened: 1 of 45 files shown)


RodZill4:
Hmm no, sorry, I don't like this coding style, even if it is perfectly valid and shorter.
I'd rather keep different styles for booleans and objects, so the code is easier to understand.

Jummit:
> Hmm no, sorry, I don't like this coding style, even if it is perfectly valid and shorter.
> I'd rather keep different styles for booleans and objects, so the code is easier to understand.

That's perfectly fine. Like I said, this is just a matter of preference.

PR #136, opened 2020-07-25
Add issue templates for easier issue reporting

Calinou:
These templates should ensure people provide the required information when reporting issues.


RodZill4:
Thanks a lot Hugo. This is exactly the kind of (much needed) things I can procrastinate forever. ;)

PR #154, opened 2020-08-24
Use the Dummy audio driver since Material Maker doesn't play any sounds

Calinou:
This prevents Godot from appearing as an application playing sound while Material Maker is running. This also decreases CPU usage (especially on macOS due to a known engine bug).


RodZill4:
Bye bye april fools features...

PR #157, opened 2020-08-28
Capture cursor on preview 3d interaction

escargot-sans-gluten:


RodZill4:
@escargot-sans-gluten I know it's not a common use case, but the behavior is weird if you hold both left and right mouse buttons.

RodZill4:
BTW, those automated builds are wonderful. I can directly test the result before merging. :)

PR #188, opened 2020-10-09
Use thumbnails instead of names in preview model selection

bonbonmiel:
I usually have a hard time switching from different preview, so I made this to avoid having to read a small text:

image

I had to use a hack to show the default cube after generating the previews so any help/feedback about that (and anything of course) is welcome.


RodZill4:
Damn, that's really nice, but we're too close to release (this week-end I hope). I think I will keep this for next release.

Would it be possible to modify the thumnail and remove that rodzlabs logo texture? I guess a uniform colored surface would be better. Custom would also deserve a different model (suzanne?).

RodZill4:
BTW, I just looked at the code... Aren't those static images?

bonbonmiel:
Thumbnails are generated on startup for each preview mesh instance (see `preview_objects.{gd,tscn}`).
I can make the changes you mentionned however for the size the problem is that they still look small on a large screen. Their size is defined in pixels and ideally I think this would be exposed in user preferences but I don't know how to do it, right now it is hardcoded.

bonbonmiel:
Okay so after experimentation I got two not-so-bad options:
thumbs

I think this looks so much better!
I personnally think texture-on-meshes is much more readable than texture-on-ground (you directly see the surface direction with curvature).
I also removed the text which looked bad.
Any feedback?

RodZill4:
Hmm if thumbnails are not animated, I think they should be integrated as png images.
Of course we should keep the scene/code that creates them so we can generate more thumbnails if we integrate new meshes.

RodZill4:
Oh and yes texture on object looks better, but maybe use a color instead of grey as background.

bonbonmiel:
This changes the thumbnail background, makes thumbs png images saved on res (thus they can only be generated from a dev project), and changes the Custom mesh to suzanne.

Currently thumbs are not showing in the builds (which then falls back to the text-version), I'm not sure why yet.

Any feedback?

bonbonmiel:
There's a `.gdignore` in there so the existence of meshes/ was hidden from me all this time...

RodZill4:
OK, I'm currently having a look at your changes. Moving some files around...
There is a weird issue, but I'm not sure it's related to your changes, must investigate.

RodZill4:
No, sorry, it seems it's fixed now.

RodZill4:
Looks like exported version works, now.

RodZill4:
@bonbonmiel , are you happy with the PR as it is now?

bonbonmiel:
I think its ok for now.
Althought there is room for ui improvement imo, it achieves its main purpose and further modifications would require more planning and discussion beforehand.

bonbonmiel:
Oh wait thumbnails for environments are actually really easy to add, do you think it could be bundled now?
image

RodZill4:
Ok

bonbonmiel:
@RodZill4 forgot to ping you back. It should be good now, if you want to review.

RodZill4:
I did notice your commit but did not have time to have a look at it yet.
Hmm do you have a Twitter account ? I spotted @bonbon_au_miel, but not sure it's you.

bonbonmiel:
@RodZill4 Yep that's me!

PR #221, opened 2020-10-29
Add zooming and panning to preview 2d

wojtekpil:


RodZill4:
Hmm looks like this one still needs a lot of work:
- the behaviour should be similar to the Reference panel (zoom around cursor, pan with middle mouse button as if you just grab and move the image)
- didn't test, but looking at the code, this breaks export and parameter controls
- there are other related requests, maybe this PR is a good excuse for them

wojtekpil:
Hmm I look into that. Good to know that reference panel has similar functionality. I don't use it since loading image is broken (at least on linux) maybe I should also look at this :) I exported it to "unreal" and looked fine so I am not sure about second point.  To be honest I was just fed up with lack of zoom on 2d nodes, so I implemented it for myself. It's really basic implementation and can be improved.

RodZill4:
The export function that this patch breaks is the one in the 2D preview's context menu.

wojtekpil:
I made few changes, it should zoom to cursor right now and pan with middle button. (I think there is a bug in reference panel - "multiplier" variable is calculated incorrectly and it doesn't work on stretched panels) I also fixed export in 2d preview's context menu.

RodZill4:
Hmm more info about the multiplier variable bug? I don't understand.

wojtekpil:
Ok so here is a problem on the video: https://streamable.com/5031j1 I think that the variable "multiplier" is incorrect but It's only a guess. I think there is a problem with conversion from "Panel space" to Shader UV space. I didn't look further into it.

RodZill4:
All this should be solved in master now (fixed it on another branch while investigating a bug).
@wojtekpil , thanks for helping, and please don't hesitate to tell me if master behaves how you expected.

wojtekpil:
Hi, yes it seems to be working correctly now. It seems that, this merge request is no longer required.

Issue #99, opened 2020-03-27
Open node search when dragging out from inputs


Jummit:
I think it would be useful if the same search that opens when dragging out from output ports opened when dragging out from input ports as well. It would connect the first output of the new node to the input.

Tar0x:
This should be closed @RodZill4

Issue #219, opened 2020-10-28
[Proposal] Quick search redesign


Tar0x:
**Feature/enhancement description:**
What about this UI redesign to make content inside quick search more readable? (font is different, because I could't match the font from Godot, even thought I used the same font). Basically the last object in the hierarchy chain would be bold and the rest would be darker with that arrow in the middle (I find it more readable than without the arrow). And also I would add like 2 px space between thumbnail and the text :) 

![image](https://user-images.githubusercontent.com/73115159/97439201-27ac7f80-1926-11eb-9bb9-03a466c3ea5d.png)

RodZill4:
Good idea, but ItemLists in Godot do not support multiple text colors per item. So this needs a new custom widget.

bonbonmiel:
I made this, it is working but it probably needs further discussion and adjustments.
Notably, some paths are too long for the popup size (this was already the case before).
Making a PR now so you can test it, any remarks?

![new](https://user-images.githubusercontent.com/71937306/98424050-cfe1e700-2090-11eb-95e2-c74f39ec062c.png)

Tar0x:
It's great <3 Maybe I would make those spaces shorter (in front and behind the arrow) and the arrow darker. It might be good to have have that pop up window a bit wider and speaking of those long paths...when you mouseover, that text could be moving (like in train stations or buses :) )

bonbonmiel:
![new3](https://user-images.githubusercontent.com/71937306/98482046-f02fb400-21fe-11eb-8479-5e04d8b4988d.png)
@Tar0x any feedback apreciated :)

There is still a question that needs to be answered: how should filtering behave when quick connecting? There are 3 possibilities:
 1. reset the filter
![reset filter](https://user-images.githubusercontent.com/71937306/98482299-cecfc780-2200-11eb-80be-bc4986cda7f6.gif)

 2. only reset the filter in the context of the current quick connection and bring it back after
![ignore filter](https://user-images.githubusercontent.com/71937306/98482740-ab5a4c00-2203-11eb-8cf7-43b1bb4b4f6f.gif)

 3. reuse it
![keep filter](https://user-images.githubusercontent.com/71937306/98482383-47cf1f00-2201-11eb-85a4-37fa47437c5a.gif)

I personally think 2. is the best option.


For showing the full paths, I think using tooltips is better that animated text :)

@RodZill4 what do you think about this?
Also, do you know if it could be possible (and a good idea) to plug node descriptions into the tooltips?

[...]

PR #230, opened 2020-11-07
Replace ItemList with custom list in add_node_popup

bonbonmiel:
#219


RodZill4:
Hmm I'm not sure I like the new behavior with the filter (the filter remains, but is not always applied when showing the popup).

RodZill4:
Hmm @bonbonmiel, thanks for that contribution, I modified a few details and added quick connect for inputs. Please tell me if you like the result (or if you find bugs I'd have introduced) so I can merge this PR.

bonbonmiel:
Hey,
 - The filter was purposely not filtering when quick-adding a node because the options have already been filtered though a compatibility pass (so I figured it was better not to automatically re-filter them through the current search).
Peek 2020-11-07 18-01

 - I'm personnally happy with the ui and the feature set but there was feedback in the issue too so maybe it could be polished a bit before merging (e.g. the filter behavior).
(also for the record, quick-connect for inputs closes #99.)

 - There is also a bug where custom nodes you just added do not show in the list before you restart the program but I guess this doesn't have anything to do with this popup.

 - Regarding paterial spray, I didn't use it seriously for now. I'll sure give feedback when I do!
image

RodZill4:
Well, that's *Material* Spray, I just typed a bit too fast (and too far from the center of keys) when commiting.

PR #235, opened 2020-11-17
Remove "A node that" from every node description

### LINKED ISSUE #232: Remove "A node that" from nodes descriptions
Most nodes descriptions begin with "A node that":
anode

I believe that this is superfluous information in the context of a node-based tool. (if it is repeated everywhere it could as well be omited)
I propose we just strip this to have descriptions like:
image
(this is the current unmodified description)
This makes more sense to me as it is more to-the-point.

skyace65:
Removes the phrase "A node that" from every node description. Closes #232.


RodZill4:
@skyace65 I hope you didn't do that manually!

skyace65:
@RodZill4 I want to say I didn't xD, but I couldn't find a way to automatically replace everything and capitalize the first letter of the next word. But doing it manually didn't take too long so it wasn't that bad.

RodZill4:
@skyace65 

 sed -i "s/A node that \(.\)/\U\1/" *.mmg

PR #244, opened 2020-12-05
Improved 3D preview

wojtekpil:
Screenshot from 2020-12-05 23-05-21
This is a draft of modifications described here: https://www.reddit.com/r/MaterialMaker/comments/k4r8nu/mm_improved_3d_viewport_proposal/

This preview currently overrides "Epping Forest" environment with new 3d preview. Still not ready to be merged, and uses FXAA of beta Godot 3.2.4 version.

Currently the list of changes include:
- new hidden HDRI, with better highlights, etc, with dark background in viewport
- higher HDRI radiance size for better quality lights
- linear to srgb colors conversion (also currently in master)
- changed camera FOV to 35
- displacement (tessellation) in custom meshes
- fixed transparency in "tessellated" meshes
- disabled backface culling
- no cracks in "tessellated" materials
- enabled shadows and screen space shadows (contact shadows)
- FXAA antialiasing


Calinou:
> enabled shadows and screen space shadows (contact shadows)

Contact shadows are known to be broken (they can cause GPU lockups), I'd disable them.

wojtekpil:
> Contact shadows are known to be broken (they can cause GPU lockups); I'd disable them. See godotengine/godot#35316.

Oh, non-working contact shadows are very bad news. I couldn't get any kind of working result with standard ones. I know that there is a problem with precision and shadows are very often in wrong place. Maybe we should increase scale of objects? It seems to help in my other experiments. 

> FXAA, while cheap, results in a loss of sharpness on textures. The problem gets worse at smaller viewport sizes. I'd make FXAA optional somehow, probably disabled by default. I think we already have MSAA enabled by default anyway.

MSAA is nice but doesn't work with transparent materials (with alpha scissors). For me it looks a lot better with FXAA than with aliased edges. 
image

There are few other changes on this image, but edges are really noticeable.  To be honest I am struggling with this problem for a long time. It's especially visible in vegetation assets. Recently I was reading about SMAA and it could probably be implemented in Godot. But maybe it's better to wait for Godot 4.0 and make FXAA optional.

Calinou:
> Oh, non-working contact shadows are very bad news. I couldn't get any kind of working result with standard ones. I know that there is a problem with precision and shadows are very often in wrong place. Maybe we should increase scale of objects? It seems to help in my other experiments.

You can decrease the Max Distance in DirectionalLight and also decrease the bias slightly.

SIsilicon:
> Make sure to test this on low-end/old GPUs, as high radiance sizes are known to crash on startup there.

I concur. The project's window currently freezes for me. By that I mean that the UI is responsive, but the display is completely broken! Setting the radiance to 128 made it with again for me.

PR #252, opened 2021-01-02
Add a Curvature Map generator

SIsilicon:
You already have an inverse map and object normal map generator. I think it would be logical to have a curvature map generator too. With this, you'll be able to create some more effects, especially when it comes to creating a "wear and tear" look for materials.
image

The implementation is a near direct port of blender's code [here](https://github.com/blender/blender/blob/594f47ecd2d5367ca936cf6fc6ec8168c2b360d0/intern/cycles/blender/blender_mesh.cpp#L541).


RodZill4:
Thanks for submitting this PR!
Hmm curvature map is on the roadmap, but I had a shader-based implementation in mind. I'm afraid your Gdscript implementation would be slow for high poly models. I'll keep it as a reference for now.

SIsilicon:
There's an issue with doing it in a shader. Wouldn't you need to be able to have access to _all_ of the vertices to get it working?

SIsilicon:
Plus, it's possible to use threads to get it running faster. But I could also look into using a shader based approach. I made an SDF generator via Godot shaders before. It was _very_ cumbersome.

RodZill4:
It would be possible, but I don't think it's a good idea to introduce C# just for this feature. 
I'll investigate the shader approach...

Calinou:
Indeed, I would prefer Material Maker not to require C# to make it easier to run from source.

SIsilicon:
What a coincidence! I'm investigating on using shaders too. So far, I believe that this should most likey be done in different stages.

Stage 1:- Sort the vertices to find duplicates easier (CPU).
Stage 2:- Find duplicates (GPU)
Stage 3:- Average out normals across duplicate vertices (GPU)
Stage 4:- Find neighboring vertices (edges) for each vertex (CPU)
Stages 5 and 6:- Calculate curvature per vertex (GPU)
Stage 7:- Render curvature map with the curvature data (GPU)

RodZill4:
Hmm seems your approach would be suited for compute shaders, but we only have vertex/fragment shaders right now.

The method I'm testing right now is entirely in 2D: using the inverse uv map and the normal map, I calculate the derivative of the normal along the tangent and cotangent, which gives the curvature relative to the U and V axes. The result is okayish (difficult to tune the amount of smoothing required).

image

The result you obtain is better, and it looks like you managed to integrate your code nicely. Maybe keep your patch as-is for now and come back to it when we have compute shaders in Godot?

SIsilicon:
Guys...
Forget the shaders...
I made a zero IQ move with my code.
Before:
image
After:
image

Dictionaries people. A good example to learn to use 'em.

SIsilicon:
A Suzanne model with 2 subdivisions (7958 verts) used to take ~_149_ seconds to generate a curvature map.
Now it only takes less than **half a second**!

SIsilicon:
Ok, here's the benchmark using Suzanne with multiple subdivisions.
subdivision 1 (2,012 verts) :- <0.1 seconds
subdivision 2 (7,958 verts) :- ~0.27 seconds
subdivision 3 (31,658 verts) :- ~1.1 seconds
subdivision 4 (126,290 verts) :- ~4.77 seconds

SIsilicon:
This is fast enough to work with, right?

RodZill4:
Yes it is fast enough. We can optimize it later if/when needed.

SIsilicon:
I'm still a little embarrassed about the initial EdgeMap implementation. 😅

Jummit:
Note that it doesn't have to be real-time, as it will be a one-time bake.

SIsilicon:
Hang on! I just said that it still shows banding artifacts.

RodZill4:
@SIsilicon Oops, sorry, did not see your comment. Hmm I'm currently merging a branch that moves the map generator files around, so please take your time. Worst case I can get the changes directly from your repository when it's ready so you don't have to worry about that. ;)

SIsilicon:
Doesn't take too long really. In fact, it's already done.
image
Left: without data packing, Right: with data packing

SIsilicon:
On an unrelated note, throughout the entire time I was implementing this feature, the project crashes consistently! I discovered that it was because of the radiance sizes of the environments in the preview 3d scene. They were too large for my PC to handle. Reducing to 128 did the trick. Can this be reduced in the project, or at least be configurable?

Issue #254, opened 2021-01-04
Add an AO and Thickness map


SIsilicon:
**Feature/enhancement description:**

For those who don't know, AO stands for ambient occlusion, so an AO map is a mesh texture that stores how occluded each part of said mesh is occluded by other geometry.
A thickness map is like an AO map, but with the normals inverted. It stores information of approximately how thick each part of a mesh is; useful for when you want to fake subsurface scattering.

While these could both be created in an external program, having a button for it along with the other baked mesh textures would improve efficiency, especially when you have a low-endish PC and want to run as little programs as possible.

I can have them implemented, and have already thought of how.
1. A BVH (Bounding Volume Hierarchy) of the mesh's triangles is created. We're gonna be ray-tracing this thing!
2. Then this BVH is transfered to a floating-point texture. We're gonna be ray-tracing on the GPU!
3. Doing it in tiles, progressively or both; we ray-trace the BVH in a shader with the inverse-UV and object normal maps too. Since recursive functions are non existent in shaders, a stack must be implemented manually.
4. We denoise the result with smart bilateral blur passes.
5. Profit!!!

SIsilicon:
Sweet, sweet progress!
I got the BVH construction working and also got BVH _traversal_ working on both CPU and GPU.
The mesh you see in the screenshot below is fully raytraced. Can you believe that? :D
![image](https://user-images.githubusercontent.com/34734122/103793238-f8934800-5011-11eb-9c16-b631a342ffb1.png)
It runs surprisingly well on my Intel GPU, even when I bump up the subdivision level a little. The next step would be to put this ray-tracer to good use, and make an AO map with it. That and optimize the BVH builder, 'cause even though building it doesn't take too long, transferring the data to an array to be transferred to an image texture does _not_ seem to scale well right now.

Tar0x:
Wow, really cool! :) It would be cool to be able to bake from highpoly model to lowpoly model. Usually you don't want to waste time unwraping highpoly model and transfer UVs from lowpoly one. I am not sure, how hard it would be to implement something like that (speaking from an artist perspecive) :) Keep it up :)

SIsilicon:
BVH :- __optimized.__
Map generation :- __underway.__
![image](https://user-images.githubusercontent.com/34734122/103828429-a02b6d00-5048-11eb-9498-044651e119a9.png)

SIsilicon:
@Tar0x What you have proposed sounds like it would come in pretty handy. :)
[...]

PR #257, opened 2021-01-08
Added AO and Thickness Map

SIsilicon:
In this pull request, two new mesh maps are available now: An AO(ambient occlusion) map, and a thickness map. This opens up new possibility for material effect, such as faking translucency, and settling dust in crevices.

**Ambient Occlusion:**
image
**Thickness:**
image
Closes: https://github.com/RodZill4/material-maker/issues/254


RodZill4:
@SIsilicon Thanks a lot for this contribution!

Can you recommend an article I can read (or whatever) to understand how this works?

SIsilicon:
Well, ngl this is pretty complicated, especially with sparse comments, but I'll point to the resources that helped me figure out how to implement this.
https://github.com/benraziel/bvh-tree
https://www.scratchapixel.com/lessons/advanced-rendering/introduction-acceleration-structure/bounding-volume-hierarchy-BVH-part1

If you have any questions or if there's a bug in the code/shader, please feel free to let me know! Next time I'll put more comments.

SIsilicon:
Other than the BVH itself, you can use the `intersects_bvh` function in shaders like a normal ray tracer. It takes the ray origin and direction, and returns the distance from the ray the intersection happens. 65536 means no intersection. it can also output the normal at the hit point, but since no normal information is in the BVH, it's generated on the fly, giving a flat shaded look if used.

PR #286, opened 2021-02-14
Don't stop the 3D preview rotation animation when zooming in or out

Calinou:
The rotation is still stopped when pressing any other mouse button.

I've often wanted to zoom while letting the automatic rotation do its job :slightly_smiling_face:


vreon:
You might be able to get away with deleting this line entirely... I'm doing that in my fork and it feels pretty usable.

It's nice to be able to manually rotate without stopping the automatic rotation (which you can already control through the menu).

PR #292, opened 2021-02-21
added a mode selector to the fill_to_uv node for square UVs

Arnklit:
Adds a drop down to Fill to UV node that allows you to choose between Stretch and Square output for the UV-islands.


RodZill4:
Hmm just had a look at the patch, didn't test, but couldn't the functions you define be global ones?

Arnklit:
Sure, I can easily make it into global functions instead. I wasn't really sure if it was better to do it one way or the other. To be honest I'm still not sure when it's best to do one or the other. Do you have any kind of rule of thumb on when you'd prefer it to be a global function as opposed to an instance one?

Issue #237, opened 2020-11-25
some brick node outputs (UV, random color) don't tile well


novemberist:
**Material Maker version:**
happens on both latest version cloned from github and official release

**OS/device including version:**
Linux
AMD RX580 with Mesa 20.2.3

**Issue description:**
As already described in this [post on reddit](https://www.reddit.com/r/MaterialMaker/comments/jpn8bm/help_needed_tileable_brick_material/) I'm having issues with the UV and random color output of the brick node. Especially when using the default offset value of 0.5 or somewhere close, the offset half-bricks on every other row (those that actually have to tile) from the left and right edges don't fit together, since they don't "wrap around" with the same color and UV. This unfortunately produces very bad results at the seams, where the material tiles. 

Since other users have reported, that they don't have this issue, I'm assuming this is some driver/GPU-specific bug. Thus I'm opening this issue here.

**Steps to reproduce:**

see above link to reddit, where the issue was described in detail

fccDunce:
I can confirm this issue. 
(Windows 10, Nvidia GTX 760)

novemberist:
I don't think this issue should be closed yet, since #293 only resolves the tiling of the random brick colors. The more severe issue with the tiling of the brick UV still persists, as you can see in the following screenshot

![Bildschirmfoto vom 2021-03-07 06-02-20](https://user-images.githubusercontent.com/7927471/110229743-e9465300-7f0b-11eb-9946-d3b55e1d1f1a.png)

novemberist:
Sorry. I'm not sure if this was accidentally or deliberately fixed in the meantime, but with the latest version from master, I can no longer reproduce this issue.

PR #293, opened 2021-02-25
changed bricks node to use center of brick for random color seed to f…

Arnklit:
The random ID's were being generated from the position of the bricks, but not their centers causing the unique colors not to tile correctly.

This should partially fix https://github.com/RodZill4/material-maker/issues/237.


novemberist:
Any chance you could apply the same fix to the UV output? 

Bildschirmfoto vom 2021-03-07 06-02-20

Arnklit:
Ah yeah I think I can easily fix that. I'll make a pull request for that a bit later today.

Arnklit:
Hmm, I cannot reproduce this error. I'm not seeing these issues on the UV layout. I remember I checked this as well when I saw this bug. Basically my fix for the Random Color was to do the same as the unique ids in the UV layout does in the blue channel as they were already using the center of the brick. Could you try to send a file where you lock off the randomness on the bricks node when it creates the error and I'll see if I get the error on the same file.

novemberist:
Sorry. I'm not sure if this was accidentally or deliberately fixed in the meantime, but with the latest version from master, I can no longer reproduce this issue.

PR #299, opened 2021-03-02
adding new smooth curvature node

Arnklit:
Adding new smooth curvature node

Example:
8r66VLZD7V


Arnklit:
@RodZill4 do you want me to close this since you updated the old curvature node now and added it to the library?

PR #301, opened 2021-03-07
Optimize PNG images losslessly

Calinou:
Command used:

    parallel oxipng --strip --zopfli ::: **/*.png


RodZill4:
@Calinou this is nice, but it never ends... Shouldn't we choose the lazy man way and add this operation to Github actions?

Calinou:
> Shouldn't we choose the lazy man way and add this operation to Github actions?

Indeed, you can add [ImgBot](https://imgbot.net/) to the repository to automatize this.

RodZill4:
I installed oxipng so I can optimize images when they are created/updated. Thanks!

PR #302, opened 2021-03-07
Add initial support for internationalization + French translation

Calinou:
## TODO

- Make more texts translatable using `tr()` calls in scripts.
  - Figure out how to translate tab names, as these are based on node names. However, we need to avoid breaking node references in scripts…
- Add language selection to Preferences.


RodZill4:
At least this translation reveals that the Parameters... dialog has a size problem. ;)

PR #304, opened 2021-03-09
adding new fbm node with simplex noise

Arnklit:
Example of usage
ktHmnXczDi


Calinou:
Should the node's title be "FBM Noise" instead of "FBM"? This would make graphs more descriptive at a glance.

Arnklit:
@Calinou sounds good to me, but I guess it up to to Rod. @RodZill4 should I change it over to FBM Noise now that I'm at it?

RodZill4:
>@Calinou sounds good to me, but I guess it up to to Rod. @RodZill4 should I change it over to FBM Noise now that I'm at it?

Sounds like a good idea.

PR #316, opened 2021-03-29
Allow painting and exporting textures in 8K resolution (8192x8192)

### LINKED ISSUE #186: Allow exporting higher texture resolutions (more than 4K)
Currently Material Maker allows to export textures with up to 4K resolution.
I wish 8K and 16K were also possible.
Is that doable? What do you think?

Calinou:
This requires a powerful GPU with at least 4 GB of VRAM (6+ GB highly recommended).

On a GeForce GTX 1080, it works successfully with most examples I tested. Of course, you'll have to be patient while 8K textures are generated and exported :slightly_smiling_face: 

This closes #186.

### DIFF
--- a/addons/material_maker/engine/gen_buffer.gd
+++ b/addons/material_maker/engine/gen_buffer.gd
@@ -40,7 +40,7 @@ func get_type_name() -> String:
 	return "Buffer"
 
 func get_parameter_defs() -> Array:
-	var parameter_defs : Array = [ { name="size", type="size", first=4, last=12, default=4 } ]
+	var parameter_defs : Array = [ { name="size", type="size", first=4, last=13, default=4 } ]
 	match version:
 		VERSION_OLD:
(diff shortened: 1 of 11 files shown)


RodZill4:
Hmm sounds cool. How's painting framerate at 8K?

Calinou:
> How's painting framerate at 8K?

I didn't test painting yet, but I'll try that. I get a feeling it might be too slow to be usable, so I can probably remove the ability to paint in 8K.

RodZill4:
> I didn't test painting yet, but I'll try that. I get a feeling it might be too slow to be usable, so I may have to remove the ability to paint in 8K.

Even if painting is near unusable, could be a good idea to keep it for procedural layers and lucky owners of ultra high end GPUs.

RodZill4:
Seems painting is completely unusable at 8K. I removed this option for now...

Issue #270, opened 2021-01-31
Either Ctrl+N or a little + tab for new materials for the lazy (or both).


some9000:
Ok, so this is incredibly lazy, but it would be nice to have just a quick click/keyboard shortcut for new material. Mostly because other editors have that. But it would be a pleasant option to have.

Here is my amazing illustration for this:
![material_maker_FhDzuKTDPV](https://user-images.githubusercontent.com/17021558/106378034-f5d60980-63a9-11eb-822c-ce4fc1daa762.png)

While we are at the interface - quitting says "Alerte !" as the window name:
![material_maker_BypU21P7Fn](https://user-images.githubusercontent.com/17021558/106378058-20c05d80-63aa-11eb-80a2-ae6b6c454416.png)

PR #319, opened 2021-04-03
Add more keyboard shortcuts

Calinou:
- <kbd>Ctrl + N</kbd> to create a new material.
- <kbd>Ctrl + Shift + N</kbd> to create a new paint project.
  - Prevent the New Paint Project dialog from being opened more than once.
- <kbd>Ctrl + Shift + Q</kbd> to close the current project.
  - <kbd>Ctrl + W</kbd> is already taken, so use the same shortcut as the Godot editor's "Quit to Project List" action.

This closes https://github.com/RodZill4/material-maker/issues/270.

### DIFF
--- a/material_maker/main_window.gd
+++ b/material_maker/main_window.gd
@@ -50,8 +50,8 @@ const RECENT_FILES_COUNT = 15
 const THEMES = [ "Dark", "Default", "Light" ]
 
 const MENU = [
-	{ menu="File", command="new_material", description="New material" },
-	{ menu="File", command="new_paint_project", description="New paint project" },
+	{ menu="File", command="new_material", shortcut="Control+N", description="New material" },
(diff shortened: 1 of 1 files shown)


RodZill4:
Hmm would there be a more general way of blocking shortcuts when a dialog is open?
Using Godot's pause mechanism maybe?
I see this problem all over MM and would like to fix it once and for all.

Calinou:
> Hmm would there be a more general way of blocking shortcuts when a dialog is open?
> Using Godot's pause mechanism maybe?
> I see this problem all over MM and would like to fix it once and for all.

I suppose we could have a variable in `main_window.gd` that becomes `true` whenever a dialog is open, and `false` when the dialog is closed. Then prevent some shortcuts from working if that variable is `true`.

PR #322, opened 2021-04-06
added random color output to color tiler node

Arnklit:
Added in the random color output by packing the tiler functions output into a mat4.

The color tiler does not have it's own page in the manual, only a mention in the Tiler page that there is a color version, so I did not make any changes to documentation.

### DIFF
--- a/addons/material_maker/nodes/tiler_color.mmg
+++ b/addons/material_maker/nodes/tiler_color.mmg
@@ -18,7 +18,7 @@
 		"ty": 4
 	},
 	"shader_model": {
-		"code": "mat4 result = tiler_$(name)($uv, vec2($tx, $ty), int($overlap), vec2(float($seed)));",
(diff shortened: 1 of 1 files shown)


RodZill4:
Uh? Why use a mat4 instead of an output parameter?

Arnklit:
Sorry I don't follow. I probably did it in a stupid way. I thought this would be the only way to get more that 4 channels of data back from the function. Otherwise the tiler logic would have to run twice right?

Arnklit:
or can you just assign to a variable in main code while you are inside an instance function?

RodZill4:
No, glsl functions accept output parameters. Just add one and pass a variable defined in the Code tab.

Arnklit:
ohhhh. Wow that makes a lot of things easier. I'll give that a go.

RodZill4:
Looking at the code, I believe there's still a problem. You can probably trigger it by blending 2 Tiler nodes.

All variables declared in the "Main code" tab must be unique (they are defined for each usage of each node). Prefixing their name with "$(name_uv)" should fix this.

Arnklit:
Ah of course. Added that now. Thanks for taking the time to give me some feedback on this so I know for next time,

PR #325, opened 2021-04-08
Allow drag-and-drop rearranging tabs and always display the Close button

Calinou:
Always displaying the Close button avoids layout shifts while changing tabs, which can have unexpected results when clicking around (such as accidentally closing tabs).

### DIFF
--- a/material_maker/main_window.tscn
+++ b/material_maker/main_window.tscn
@@ -210,7 +210,8 @@ __meta__ = {
 margin_right = 950.0
 margin_bottom = 24.0
 tab_align = 0
-tab_close_display_policy = 1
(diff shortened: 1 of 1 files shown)


RodZill4:
Hmm rearranging tabs does not seem to work well. I don't manage to drag a tab from all the way left to all the way right with only 3 tabs...

Calinou:
> I tested a dev build (based on 3.2.3). Did you test this with a 3.3 RC?

I think I tested this on 3.2.3, but I don't remember it.

Calinou:
Could this be merged to resolve https://github.com/RodZill4/material-maker/issues/359? Thanks in advance :slightly_smiling_face:

PR #348, opened 2021-08-06
Add achievements to help users discover features

RodZill4:


RodZill4:
Here's how to help with this PR:
- Achievements are defined in the material_maker/tools/achievements/achievements.gd file
- Each achievement is unlocked using the following code:

```
var achievements = get_node("/root/MainWindow/Achievements")
if achievements != null:
    achievements.unlock("achievement_id")

```

PR #351, opened 2021-08-07
Makes curve editor resizeable and adds curve presets

Arnklit:
This adds a bit more functionality to the curve editor. I think it's really nice to have it resizable, since it's much easier to make a nice curve  if you make the popup square. Then I added the presets and an invert button. I hardcoded the presets into a script, maybe it would be better to save the presets as curve resourses and read them in and convert them to MMCurves?

The icons for the presets have dark and light versions and load based on current theme, but I'm not sure if it's a very good way I do that, maybe you could check if it could be done better and let me know.

Before;
material_maker_eE04wKOjCu

After:
Dhag78OkNU


RodZill4:
This deserves a doc update... :D

PR #358, opened 2021-08-30
Chocolatey installation method added

jakublevy:
Hi,
I created and maintain a Chocolatey package, would be nice, if you could add Chocolatey as an installation method.

Thanks


RodZill4:
Oops, edited the README directly, without looking at the PR (saw it in my mailbox).

PR #364, opened 2021-10-07
Add graph nodes that copy contents from other graph nodes

easynam:
(WIP)

It would be useful to be able to instance subgraphs so that you can edit a graphs in one place and have that edit take effect in multiple places.

This pull request currently adds a very hacky way to achieve something similar:
 - adds a Regenerate Copies button to the UI
 - clicking this button finds all subgraphs with names like copy_X, copy_Y etc. and then finds graphs named X and Y and copies their contents to the respective copy graphs.

Why this is currently bad:
 - it relies on names to add behaviour - a proper implementation would use a node that can reference another node or something 
 - performance:
   - replacing the full contents of each graph is very bad since even small changes result in everything having to be recalculated (though good solutions such as diffing would make the implementation much more complicated)
   - serialising and deserialising the graphs freezes the program when there are a lot of nodes to copy
 - changes to parameters and inputs/outputs don't currently update on the UI straight away after being copied (wasn't sure how to update properly)
 - changing parameters also sets all their values to the values from the source graph - it would be nice if parameters that already exist and have the right types keep their existing values
 - have to manually update
 - only works with other graph nodes on the same level

### DIFF
--- a/material_maker/doc/subgraph_nodes.rst
+++ b/material_maker/doc/subgraph_nodes.rst
@@ -10,8 +10,10 @@ This will replace all selected nodes with a single subgraph node, without
 modifying the overall material description.
 
 When the subgraph is created, it is shown in the graph editor, and can be
-renamed using the top right text field. the yellow doc button next to this
-text field can be used to define the Group Node's tooltip.
+renamed using the top right text field (which is recommended if you intend
+to use several instances of that subgraph in your project).
+The yellow doc button next to this text field can be used to define the
+Group Node's tooltip.
 
 The subgraph node is not editable by default, but using the **Control+W**
 shortcut will make it possible to edit its contents using the pencil button.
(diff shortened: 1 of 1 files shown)


easynam:
Theres some diff code now to reduce the amount of needless buffer recalculation - it definitely needs some cleanup but also its 1am.

Also haven't tested if it works in all situations, though it seems to perform ok on the pretty big graphs i threw at it.
the diff is currently slow but still a lot faster than recalculating buffers used to be

RodZill4:
OK, thought about this a bit more.

Let's introduce this feature in 2 steps.
- (maybe possible for 0.97) just modify the UI and keep the feature as-is ("limited" to mirroring graph nodes). There's no real need for improving the diff/patch system for this, just make sure it works, and detect correctly if mirroring is possible or not (i.e. mirrored nodes connections compatibility).
- generalize the diff/patch system so it supports all nodes, so they can be applied to whole materials. This will be useful for undo/redo and generalize this feature to all node types.

PR #369, opened 2021-10-30
Undo Redo support

metincetin:
Added an extensible undo/redo system that currently works with delete, move, connect/disconnect with some issues needing to be addressed.
# How it works
Every action is a UndoAction, which extends Resource. Any type of action that needs to be implemented, can be written as a class that extends UndoAction.
UndoAction contains do, undo, and destroy methods. When an action is performed (moving nodes, for example) a MoveNodeUndoAction is created and added to the stack. When undo command is run, UndoRedo class calls `do` method on the current action, then goes back. For redo, does the same by calling the `undo` method (possible bad naming here)

## Example Usage
```gdscript
func disconnect_node(...):
    ...
    var action = DisconnectNodeUndoAction.new()
    action.graph_edit = self
    action.from = from
    action.from_slot = from_slot
    action.to = to
    action.to_slot = to_slot
    $UndoRedo.add_action(action)
```

- Had to alter `remove_node` function, this might need reviewing. I had to create an alternative function named `destroy_node` When a node is removed, it is not destroyed until `destroy_node` is called, which is done on  `RemoveNodeUndoAction.destroy` method, when it cannot be undone any longer.
- Also had to alter parameter_changed event so that it sends its previous value. 

# TODO

- Limit the number of actions that can be undone
- Merging similar actions to one composite action (e.g. moving the same node, removing nodes with disconnecting connections)
- Issues mentioned below

# Issues

- parameter_changed event is sent multiple times when the user changes a value. I could not find any event that is sent when parameter alteration is complete.
- When a node is removed, its connections are also removed and these operations are sent as separate undo actions. This works, but requires multiple redos to revert to its original state. CompositeUndoAction, which merges multiple actions into one, and added `UndoRedo.begin_composite` and `UndoRedo.end_composite` methods to work around this, but didn't spend much time on that yet. The idea is to call begin_composite when `remove_node` function is called, and call end_composite when at the end. When `begin_composite` method is called, every action is added as a unified action in CompositeUndoAction, so when undone, everything inside this object is undone all together

**Note:** I did not realize there already is a pull request for undo/redo '#349' until this was ready to be sent as a pull request. If it works fine, can be merged. If not, totally fine to ignore it, or can be used to get ideas.

### DIFF
--- a/addons/material_maker/engine/gen_base.gd
+++ b/addons/material_maker/engine/gen_base.gd
@@ -152,7 +152,7 @@ class CustomGradientSorter:
 func set_parameter(n : String, v) -> void:
 	var old_value = parameters[n] if parameters.has(n) else null
 	parameters[n] = v
-	emit_signal("parameter_changed", n, v)
+	emit_signal("parameter_changed", n, v, old_value)
 	if is_inside_tree():
 		var parameter_def : Dictionary = get_parameter_def(n)
 		if parameter_def.has("type"):
--- a/material_maker/nodes/generic/generic.gd
+++ b/material_maker/nodes/generic/generic.gd
@@ -59,7 +59,7 @@ static func update_control_from_parameter(parameter_controls : Dictionary, p : S
 		else:
 			print("unsupported widget "+str(o))
 
-func on_parameter_changed(p : String, v) -> void:
+func on_parameter_changed(p : String, v, o) -> void:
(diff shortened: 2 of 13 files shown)


RodZill4:
Yes, I'm already working on undo/redo (I guess my PR can undo/redo a bit more actions and has merging). I'll have a look at your code anyway, could give me ideas. Thanks!

PR #376, opened 2021-11-11
Library colored nodes

Arnklit:
Note: This PR is not ready to merge

image

This an initial implementation of having the nodes get coloured based on what library section they come from.

Note: currently this only works correctly on Dark theme. It should work fine on default and light theme, once those themes get edited to have the correct defaults. However a restart of MM would currently be needed after changing the theme. I need to set it up so the GraphNode themes get remade and the graph redrawn on theme change.

The biggest issue currently is that the section a node belongs to is not part of the node definition, so I currently have to look up the name of a node in the library and read it's section from there. This is inconsistent and doesn't work well with user created nodes.


RodZill4:
I'm not sure what the right approach for defining colors (or a color identifier corresponding to sections) for nodes is. There are several solutions:

- make color part of the node's definition. I'm not sure it is a good idea because this means updating many nodes whenever we want to reorganize the library (which should not happen that often, anyway). This also makes color an attribute of the node itself (in the "engine" part of MM) which does not really make sense.
- have a "node color registry" somewhere in the UI part of MM. This could very well be in the node factory and node colors "decided" when loading libraries. That way, if one day we want to implement user defined library sections, we won't have to change anything.

Arnklit:
This should be ready to merge from my point of view. But you might want to review the change I did to how selected node are displayed. I used the trick pycbouh suggested. See the change in base.gd (end of draw() function).

The reason I did that is so we didn't have to use a drop shadow as the outline for selected nodes, which caused the nodes to change colour somewhat when selected since the drop shadow was visible through the semi transparent node. It also means we can have both an outline and a dropshadow, when needed.

I think we should merge this PR and rebase the reroute PR, so we can ensure it works correctly with the selection and don't have to merge themes it a bad way.

DARK
image

DEFAULT:
image

LIGHT:
image

PR #380, opened 2021-11-14
Add Palettize node

inforthecloud:
I have introduced _post_process() function in MMGenBuffer class which is called upon new texture is generated. In the derived class MMGenPalettize this function implements pixel by pixel replacement borrowed from Slade editor.


RodZill4:
A shader-based palettize node has been added.

PR #387, opened 2021-12-01
Binary Smooth Node

Arnklit:
This adds the Binary Smooth Node

gwjjaH9ph5

Note: This node was made by Pavel Oliva (Tarox), but he asked me to do the PR for it.


Calinou:
Does this node have a way to prevent aliasing, or should this be handled by the user using a slight blur filter or the Supersample node?

Arnklit:
> 

So this is just a step based on a blur, but it actually doesn't output buffered. So I might actually want to change the description, the resolution doesn't actually set the output, it just is what quality blur it uses.

Just a 64x64 res blur gives a lot of detail.
I28RRPbySk

RodZill4:
A bevel parameter could be exposed (width from the output step node)

Arnklit:
> A bevel parameter could be exposed (width from the output step node)

yeah I guess, but I feel like the point of the node is that you put in a hard shape and get another one out but with it's shape rounded. And the thing is the "bevel" you would be getting out would not be a perfectly linear gradient that you could easily use for a bevel but rather a selection of a smoothstepped gradient generated by the blur.

RodZill4:
> > A bevel parameter could be exposed (width from the output step node)
> yeah I guess, but I feel like the point of the node is that you put in a hard shape and get another one out but with it's shape rounded. And the thing is the "bevel" you would be getting out would not be a perfectly linear gradient that you could easily use for a bevel but rather a selection of a smoothstepped gradient generated by the blur.

Sure, but you'll often use this node for  heightmaps to generate normals, and a sharp output will generally be unusable. Sure the bevel won't be homogeneous, but it has no GPU cost.

PR #397, opened 2021-12-22
Add export support for Qt Quick 3D

nezticle:
This will export PrincipledMaterial components compatible with Qt 6.2
and greater.


RodZill4:
Nice to see you finally created a PR for this. :)
I'll have a look at it before merging, and will probably include the template file in the material node...

Issue #305, opened 2021-03-10
Baking AO map takes in account directional light probably


Tar0x:
**Material Maker version:**

**OS/device including version:**

Win 10, GTX 1050

**Issue description:**

When baking AO map, there is some kind of direction taken into account, so on one side there's not any AO and on the other side there is some kind of cast shadow.

Baking setup: 
ray count: 64
ray lenght: 128
denoise radius: 3

**Steps to reproduce:**
In the following video I've baked 2 AO maps using the same meshes, except one of them was turned 90 degree along vertical axis. https://youtu.be/sDqzOLX863c

Calinou:
cc @SIsilicon

SIsilicon:
I'll take a look at it.

Tar0x:
Sure, here are the meshes (I've added also those baked AO maps). https://mega.nz/file/Z1Uz3QBB#i3G_W1fCNnylYeIx5FMIENPQ7SUobVZ8A0g4lyjoXaU
This is not problem only for this model. I've had also the same problem on a different mesh, but it was rotated "correcly", so the shadow seemed natural to me :) @SIsilicon

PR #400, opened 2021-12-22
fix ao & thickness directionality problem

wojtekpil:
Fixes issue with directionality of ao & thickness baking. Mesh in different rotation would produce different baking result.
The issue was an assumption that fract() in glsl (inside rand() function) is mathematically correct, where in reality its calculated as x-floor(x). This results in values in range 0..1 not -1..1

### DIFF
--- a/material_maker/tools/map_renderer/map_renderer.tscn
+++ b/material_maker/tools/map_renderer/map_renderer.tscn
@@ -327,8 +327,8 @@ float intersects_bvh(vec3 ray_start, vec3 ray_dir, out vec3 normal_hit) {
 vec3 random_hemi_point(vec3 rand, vec3 norm) {
 	const float PI = 3.141592653;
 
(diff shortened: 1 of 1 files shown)


RodZill4:
Why didn't you just fix the rand function (remove the fract operation)?

wojtekpil:
Ah, good point! I am not sure if that really matter, but fract gives a range of [0..1), so after rescaling it will give values from [-1,1). Sine would give a range of [-1,1].  I tried to be compatible with the old comment in source. But yeah probably it doesn't matter.

RodZill4:
I don't think it matters, and removing fract in rand() will make the shader a bit faster.

wojtekpil:
I tried to change it but now I think it should stay as it is. We probably didn't notice brackets order in rand() function. fract() is necessary to keep everything in 0..1 range. Otherwise its [- 43758.5453,  43758.5453]. And changing that would result in totally different random function.

wojtekpil:
It should close https://github.com/RodZill4/material-maker/issues/305

PR #419, opened 2022-02-05
32 bits viewports

RodZill4:
Use 32 bits viewports from Godot 3.5.

Beware, dev builds for this PR won't work because they still use Godot 3.4.


RodZill4:
Obsolete, now Godot 4 is available.

Issue #415, opened 2022-01-29
Fill output on new brick pattern nodes doesn't tile.


Yukwuh:
**Material Maker version:**
v0.98

**OS/device including version:**
Windows 10 64 bit

**Issue description:**
The previous versions of the brick node had outputs that returned the equivalent of a fill companion node's output, the new bricks node has a fill output. However, running the fill output in to an fill companion node returns a result that doesn't tile like the previous version.
***Old Output***
![image](https://user-images.githubusercontent.com/18547888/151676277-0b038449-2df0-435d-80a2-7ba4bed0dcd0.png)
***New Output***
![image](https://user-images.githubusercontent.com/18547888/151676285-bb6c0a67-5bad-4a28-a200-17213a76bf36.png)

**Steps to reproduce:**
Connect the Brick node's fill output to a Fill-To-UV node. 
![image](https://user-images.githubusercontent.com/18547888/151676326-50fb1652-93ed-4ce3-aea0-050fa68e6ef4.png)

RodZill4:
Fixed in 0.99

PR #437, opened 2022-03-22
Ensure that fill outputs correctly tile on Bricks and Uneven Bricks 2

Arnklit:
As discussed this fixes the tiling on the Fill outputs.


Arnklit:
This resolves https://github.com/RodZill4/material-maker/issues/415

Issue #97, opened 2020-03-17
[Proposal] Add function to lock the rotation axis in 3D view


varkatope:
It's a bit harder than it should be to manipulate the 3D views and it would be nice to be able to just lock the rotation axis so that the view can only be rotated horizontally, for example, at least while using the regular left-mouse button to rotate.

Calinou:
I think at least one of the rotation axes should be locked permanently, like in Godot or Blender when you use the default orbit settings.

Tar0x:
Omg yes please! To rotate camera and then roll it is really time consuming.

Calinou:
I tried to implement this a while ago but didn't succeed in doing so. It can be done easily by using a "pivot" yaw node and a pitch node as a child, but that's the hacky way of doing it…

PR #455, opened 2022-05-08
Add different 3D navigation styles

bliepp:
**Problem description**:
As every 3D software uses it's own navigation style in 3D I find it hard to transition from one software to another as I always have to rethink navigation in 3D space. I therefore thought it would be nice to let users choose which way to navigate the 3D viewport by choosing navigation styles. Right now the default ist something similar to (but not the same as) Blenders "Trackball" orbit method. Many other software - Godot included - on the other hand uses something similar to Blenders "Turntable" orbit method.

**What I implemented**:
I've implemented something similar to the "Turntable" style navigation and created a submenu under the "Environment" popup menu for changing it. I think this would be a nice addition as I would definitely prefer the turntable-styled navigation.
Moving around is similar to the default navigation style:
* Orbiting by clicking and dragging.
* Holding Cmd and dragging zooms in/out
* Holding shift and dragging rotates the object freely
* Scrolling with the mouse wheel zooms as well
* Scrolling and while Cmd changes FOV

**How it is implemented**:
As the navigation code was very long I implemented it in a strategy like pattern. The new way of implementation allows for a more modular approach as many navigation styles can be added later as the navigation and UI logic are now strictly separated. The default navigation code was moved to it's own class but remains mostly untouched.

Edit: This addresses #97 somehow


bliepp:
Oh, I forgot that it would be nice if this setting would be saved when closing Material Maker, but I could not figure out how to do this for now.

RodZill4:
Nice! I'll have a look at this when I'm done with EasySDF update...

bliepp:
I would also think turntable navigation should be the default as it is way more common in other programs and also more intuitive.

bliepp:
Made some final touches to this PR:
* Resolved merge conflicts
* Renamed a few things
* Save this new setting when restarting MM

It is now ready for merge

bliepp:
Completed the Godot4 port of this PR and resolved all merge conflicts that occurred because of that, so it's now fully compatible again.

Once again, it's ready to merge.

bliepp:
Seems not to be needed anymore as of the upcoming 1.4 release. Thanks for the improvements!

Issue #442, opened 2022-03-31
Add a small grey outline around gradients


Arnklit:
Gradients that are black at either end, fade into the node panel colour. It might be good to add a 1px grey color around the gradient.

![image](https://user-images.githubusercontent.com/4955051/161000906-2f03be51-52d4-4107-8c3d-c55a90fcb31a.png)

I can give it a go to add this in myself, I just don't have time to try now, so I thought I'd log a bug for it.

bliepp:
I quickly submitted one since I was exploring the nodes anyway. #457

PR #457, opened 2022-05-08
Add neutral grey outline around gradients

bliepp:
Added a neutral grey outline around gradients in nodes for better visibility.
Addressing #442
export


Arnklit:
I think this broke double clicking on the gradient to open the bigger preview!

bliepp:
> I think this broke double clicking on the gradient to open the bigger preview!

I think you are right. I'll have a look at it.

Issue #451, opened 2022-05-04
Crash After Canceled Add Parameter


Waffle1434:
**Material Maker version:** 0.99

**OS/device including version:** Windows 10 Latest

**Issue description:**
After cancelling the "Add linked control" or "Add configurations" operation in a node group, a new "Unnamed" parameter is created in a broken state that will instantly crash if you attempt to link a parameter to it. However, clicking "Move parameter up / down" will fix this broken state and prevent a crash.
![image](https://user-images.githubusercontent.com/8021358/166832561-e7d747d0-01fc-4ff9-af11-02b874ae5ee1.png)

**Steps to reproduce:**
1. Create group `Ctrl+G` or open an existing
2. Parameters -> "Add linked control" or "Add configurations"
3. Right or middle mouse click anywhere to cancel the operation
4. On the newly created "Unnamed" parameter -> "Link another parameter"
5. Instant crash

meywue:
Hey @Waffle1434 

In case you cloned the project instead of using the precompiled executable, you can add the following lines to the link.gd file at line 79 to fix the problem.

`
else:
    node.generator.remove_parameter(param_name)
`

This will remove the control again once you hit MMB/RMB.
In my opinion there's a small UX problem though. If there's and existing link, you would like to update it and hit MMB/RMB, it will remove the control. But this also happened before when hitting the escape button on your keyboard.

Full file (link.gd):
`
extends Control
class_name MMNodeLink

var end
var source = null
var target = null

var node = null
var param_name : String = ""
var creating : bool = false

func _init(parent) -> void:
	size_flags_horizontal = SIZE_EXPAND_FILL
	size_flags_vertical = SIZE_EXPAND_FILL
	rect_size = parent.rect_size
	rect_clip_content = true
	parent.add_child(self)

func pick(s, n, pn : String, c : bool = false) -> void:
	source = s
	end = get_global_transform().xform_inv(source.get_global_transform().xform(0.5*source.rect_size))
	node = n
	param_name = pn
	creating = c
	set_process_input(true)

func show_link(s, t) -> void:
	set_process_input(false)
	mouse_filter = Control.MOUSE_FILTER_IGNORE
	source = s
	target = t
	update()

func closest(rect, point) -> Vector2:
	return Vector2(max(rect.position.x, min(rect.end.x, point.x)), max(rect.position.y, min(rect.end.y, point.y)))

func find_control(gp) -> Dictionary:
	for c in get_parent().get_children():
[...]

PR #458, opened 2022-05-17
Remove node's control when aborting link creation

meywue:
Fixes [Crash After Canceled Add Parameter #451](https://github.com/RodZill4/material-maker/issues/451)

The program crashes due to the link object trying to call an function of a non-existing object. This object is supposed to be set by the node the link is coming from.

Removes the control from the node when the link creation has been aborted.
The same thing happens when the user hits the `escape` key on their keyboard.


RodZill4:
Thanks a lot for this PR! There were quite a few problems in that code, so I tried to fix them at the same time (on master). please don't hesitate to tell me if behavior is what you expect (it's a little bit different from what you implemented, you can zoom in/out while holding the LMB so you can connect far from the Remote node).

RodZill4:
Already fixed on master with other Remote issues.

PR #459, opened 2022-05-27
Comment node auto sizing to selected nodes

Zhibade:
## Overview
After working on my first material using Material Maker, I found that relying on the comment node manual resizing and placement can get somewhat time consuming, especially when reorganizing nodes in more complicated graphs. This PR adds auto-sizing feature to the comment node if other nodes are selected when the comment node is created. It makes it work just like Substance Designer. 

I also added a checkbox in a new "Graph" tab in the preferences popup, in case some people would like to turn off this behavior.

## Implementation Details
During initialization in **set_generator**, if the config setting is enabled, it checks if any nodes are selected and resizes the comment node to match the position and size of the selected nodes.

I didn't want to bloat the preferences with unnecessary settings, so I hard-coded the extra padding value of the node size and position to something that I thought looked good.

## Comments
Although I am not checking if the node is created by the user or is being loaded from a saved file, every time a new graph is loaded the selection is cleared, so it doesn't cause any problems. I am not familiar with the codebase, so I didn't want to do any refactors or big changes, but please let me know if you have any suggestions.

Thank you for creating Material Maker by the way! Great free alternative to Substance Designer and similar software.

CommentNodeAutoSizing


RodZill4:
Excellent. I'll review and merge it for 1.0.

PR #467, opened 2022-06-13
Fix dilate node at 1k and higher resolutions

Arnklit:
This makes dilate fill correctly tile at 1k and above resolutions by packing the data before the buffer and unpacking after. 

Note: I also changed the default resolution of dilate to 2k. This seems to be a sweet spot for a lot of people to work in at the moment and a lot of our newer nodes default to that resolution.

Thanks to @wojtekpil for the logic of packing and unpacking.


RodZill4:
Seems to work fine. :)
Maybe we should define split/join functions in a node (that just define those functions) and insert them directly into passes. That way they can be used in several nodes, and if we come up with better split/merge we just have to update the functions.

Arnklit:
Thanks Rod, Seems like a great idea to have it available as functions.

Issue #483, opened 2022-07-10
Node preview image displays incorrectly if cursor leaves node too quickly


BlacRyu:
**Material Maker version:**
latest dev (https://github.com/RodZill4/material-maker/commit/4a6b82e3c719d603aa56e334ecfe74117e06f0de)

**OS/device including version:**
Windows 10 Pro x64
Version		21H2
OS build		19044.1766
Experience	Windows Feature Experience Pack 120.2212.4180.0

**Issue description:**
If you move the cursor off of a node immediately after maximizing the node, the program will attempt to show the node's preview image but it will not be sized correctly. The issue corrects itself after a split second (when the preview would normally become visible).

**Steps to reproduce:**
1. Right click a node's output pin to enable the preview image.
2. Click the square icon at the top of the node to minimize it.
3. Click the square icon again to maximize the node.
4. Immediately move your cursor out of the node.

**Screenshot:**
![image](https://user-images.githubusercontent.com/10762433/178161954-635e46f4-960c-4b02-b03d-6afad563db68.png)

Issue #482, opened 2022-07-10
Node preview doesn't toggle if cursor enters an overlapping element before entering/exiting previewed node


BlacRyu:
**Material Maker version:**
latest dev (https://github.com/RodZill4/material-maker/commit/4a6b82e3c719d603aa56e334ecfe74117e06f0de)

**OS/device including version:**
Windows 10 Pro x64
Version		21H2
OS build		19044.1766
Experience	Windows Feature Experience Pack 120.2212.4180.0

**Issue description:**
It seems that the preview image shown on individual nodes is meant to be visible when the cursor is not hovering over the node, and disappear when the cursor hovers over the node. Currently this behavior breaks if the cursor hovers over any other GUI element that is overlapping the node before the cursor transitions into or out of the node. This can include other nodes or UI elements that cover the node, or even sub-elements of the same node such as text input boxes, input/output pins, or the preview image itself.

**Steps to reproduce:**
1. Right click on a node's output pin to enable the node preview.
2. Move the cursor down over the node so that the preview disappears.
3. Move the cursor back over the output pin.
4. Move the cursor to the right so that it exits the node.

**Video:**

https://user-images.githubusercontent.com/10762433/178161690-4f99b972-8588-4e62-b0fc-0b7b3f98cea4.mp4

PR #484, opened 2022-07-10
Fixes issues where node preview does not toggle correctly

### LINKED ISSUE #481: Node previews don't auto-hide if already enabled when material was loaded
**Material Maker version:**
latest dev (https://github.com/RodZill4/material-maker/commit/4a6b82e3c719d603aa56e334ecfe74117e06f0de)

**OS/device including version:**
Windows 10 Pro x64
Version		21H2
OS build		19044.1766
Experience	Windows Feature Experience Pack 120.2212.4180.0

**Issue description:**
Nodes with a previewed output enabled will not toggle the previews visibility on mouse-over if the preview was already enabled when loading the material.

**Steps to reproduce:**
1. Enable a preview by right-clicking the output pin of a node.
2. Save the material.
3. Load the material.
4. Move the cursor over the node.

**Video:**

https://user-images.githubusercontent.com/10762433/178157390-35b4cde1-7fda-4600-9faf-51e4c4d03076.mp4

BlacRyu:
Fixes #481, #482,  and #483

I've tested this on various nodes, trying to make sure it would work cleanly with any type of parameter. This is my first contribution, so apologies if there are any conflicts with design or code style.

### DIFF
--- a/material_maker/nodes/generic/generic.gd
+++ b/material_maker/nodes/generic/generic.gd
@@ -1,6 +1,7 @@
 extends MMGraphNodeBase
 class_name MMGraphNodeGeneric
 
+
 var controls = {}
 var ignore_parameter_change = ""
 var output_count = 0
@@ -534,7 +535,7 @@ func preview_connect(node : Control) -> void:
 			node.connect("mouse_exited", self, "on_mouse_exited")
 		for child in node.get_children():
 			preview_connect(child)
-	
+
 func preview_disconnect(node : Control) -> void:
 	if node is Control and node.is_connected("mouse_entered", self, "on_mouse_entered"):
 		if node is Popup:
@@ -552,8 +553,11 @@ func on_mouse_entered():
 
 func on_mouse_exited():
 	if !generator.minimized:
+		var preview_parent = preview.get_parent()
+		# Fake move of preview in hierarchy, so it's shown in front of whatever
+		# control that has been created recently
+		preview_parent.move_child(preview, preview_parent.get_child_count()-1)
 		preview.visible = true
-	
 
 func update_from_locale() -> void:
 	update_title()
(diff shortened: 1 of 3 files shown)


RodZill4:
BTW, do you have this problem that when you create a new gradient cursor, it is shown in front of the preview?

BlacRyu:
> This looks pretty good, but I really don't like the gradient editor part (looks too specific, this hints it might break if value edit controls are modified).
Yeah, ideally I was hoping Godot would have a signal indicating that a new node was added to the scene tree. Then I could just check for any new children down the tree in generic.gd rather than needing to manually put checks in the control script(s). Not sure of a clean way to future proof it without something generic like that though.

> BTW, do you have this problem that when you create a new gradient cursor, it is shown in front of the preview?
I didn't notice that, no. Thanks for fixing it. I must have always created the cursors before the preview in my testing.

BlacRyu:
Actually it looks like the scene tree does emit [a signal](https://docs.godotengine.org/en/stable/classes/class_scenetree.html?highlight=node_added#class-scenetree-signal-node-added) when nodes are added, I'll see if I can use that here to make it connect new children in a clean generic way.

RodZill4:
I don't think it's a good idea, the scene tree emits a signal *whenever* a node is added, I don't think we want all nodes to listen to that signal. But maybe I'm wrong...

BlacRyu:
Yeah I'm not knowledgeable enough about the performance overhead of signals, so if that's a concern I can just leave it as it is.

BlacRyu:
It looks like the signals we really want, `child_entered_tree` and `child_exited_tree` have been added to godot 4, but aren't available in 3.x (yet). It looks like they may be added in 3.5 though: https://github.com/godotengine/godot/pull/57541

BlacRyu:
Yeah in the thread I posted above they talk about `node_added` having "a huge performance cost and is only really relevant for debugging / editor tools". When they talk about performance they may be talking more of real-time performance for things like games though which need the highest framerate possible. If you can add a few hundred nodes all with previews on and not chug it's probably fine.

BlacRyu:
Btw, Github only auto-linked the first issue I mentioned in the PR and it looks like I can't link the others. If you could, link #482 and #483 as this pr will close them.

PR #493, opened 2022-07-27
feat: spiral gradient

Theaninova:
anim

I'm new to contributing to this project, is there other things I need to add apart from the node?


RodZill4:
Hi, and thanks for contributing to this project. :)

There are a couple extra things when adding a node:
- add it to the base library, so it can be easily accessed
- add documentation for the node
- add your contribution to the About dialog and the change to changelog (I'll do those before merging)

Please don't hesitate to ask if you need help.

Hmm maybe it does not need the gradient parameter, and could output a greyscale value (it's easy to use a Colorize node if needed).
I'm not sure about the cutoffs either. Being able to select the inner/outer values as well as a bevel (for smooth transition between the spiral and uniform areas) would be nice, but that would make the node a bit too complex.
What do you think?

Theaninova:
I'll add the docs (I'm assuming they go [here](https://github.com/RodZill4/material-maker/tree/master/material_maker/doc)?), but I haven't found yet where the base library is configured...

With the gradient input I was doing it for consistency with the other gradient nodes, that all have an input like that, so I think it's reasonable to leave it there.

Theaninova:
image

I've done a few changes to get the node more in line with other gradient nodes too

RodZill4:
Nice! It's simpler now.

Theaninova:
Okay, I believe everything should be added correctly now!
image

RodZill4:
Yes, it looks perfect.
Please don't worry if I don't merge it right now. I generally merge new nodes just before release to avoid that users who run dev builds post materials that rely on new stuff (therefore not compatible with current release) on the website. So this will be merged in September.

PR #498, opened 2022-08-08
Add vec3 math operations

williamchange:
This adds additional vector math operations for the vec3 math node:
Length, Distance, Dot Product, Cross Product and Normalize

I am following the convention from the current node so it defaults to the last operation, though should it be following the regular math node which selects the first operation by default instead?

image
image


williamchange:
Added sign and modulo functions
sign_mod_sample

PR #500, opened 2022-08-10
Add sixteen segment display node

williamchange:
This adds a sixteen segment display node, which supports displaying characters from ASCII (32-125, 31 reserved to display all segments), with variable weight and rotatable diagonals.

xvisd_demo

The following characters are currently supported:

characters


williamchange:
The node now includes a `Custom` display mode(switchable between `Custom` and `ASCII`) which takes in an unsigned 16-bit float (0-65535) to determine which segments to be turned on based on bit positions:

node_preview

custom_mode_example

For example, to display `M`:
**Binary**: `0001 0010 0001 1011`
**Decimal**: `4635`
I have also written a simple [web app](https://williamchange.github.io/xvisd/) as a reference point for the segment mapping/corresponding hex/dec/bin configurations:

M_example

Alternatively, here's how the bit positions are mapped to the segments:

```
  _10__ __14__
 | \   |   /  |
 12 13 11 16  15
 |    \|/     |
  __1__ __6___
 |    /|\     |
 4  5  2  9   7
 | /   |   \  |
  __3__|__8___

 1  4    8   12    16
 |  |    |    |    |        
 0000 0000 0000 0000
```

PR #505, opened 2022-08-20
Remove redundant entry in nodes_simple.rst

williamchange:
This removes the redundant entry in `nodes_simple.rst`:

`node_simple_seven_segment_display`


RodZill4:
Looks like it's fixed already (probably while merging other PRs). Thanks anyway!

PR #510, opened 2022-08-29
Don't stop automatic rotation when rotating Preview3D

vreon:
This has come up on Discord a couple of times ([1](https://discord.com/channels/784201835334074389/789821738669572137/1013547310950662305), [2](https://discord.com/channels/784201835334074389/789948291517382759/1007279107492548618)) and I happened to have a local patch for this already :smile:


bliepp:
Actually, I think it is a good think that autorotation stops. When navigating the 3D preview one is usually aiming for a specific perspective. If the rotation didn't stop the preview would continuously change which is not what one would expect when explicitly navigating the camera to a different position.

vreon:
Closing in favor of #510.

PR #512, opened 2022-09-01
Add Triangle Voronoi node

williamchange:
This adds a new voronoi texture node which uses the triangle distance metric for cell distances to border. Based on [the shadertoy](https://www.shadertoy.com/view/ss3fW4) by [Thomas Hooper](https://twitter.com/tdhooper/status/1542214131524173825). Sample outputs:

tri_voronoi_sample mp4


axredneck:
Why does it have circular gradients if it uses triangle distance?

PR #514, opened 2022-09-06
Add Japanese Glyphs node

williamchange:
This adds the "Japanese Glyphs" node, which currently supports the 46 base characters for both writing systems in Hiragana/Katakana (excluding the Dakuten/Handakuten variants - i.e. か/が は/ぽ):

jp_glyph_sample

**Hiragana Samples**:
hiragana_samples

**Katakana Samples**:
katakana_samples


Zireael07:
Can this only produce real glyphs or could it be tweaked to produce 'fake' characters?

williamchange:
> Can this only produce real glyphs or could it be tweaked to produce 'fake' characters?

Do you have an example what do you mean by 'fake' characters? Right now this only draw the the characters as it is.

Zireael07:
https://www.shadertoy.com/view/Xtcczj (this is more for the kanji than katakana/hiragana, but I guess something similar could be used)

williamchange:
> https://www.shadertoy.com/view/Xtcczj (this is more for the kanji than katakana/hiragana, but I guess something similar could be used)

Oh cool, though I guess this is closer to what the runes node (i.e. https://www.shadertoy.com/view/MsXSRn) is doing as this produces fixed glyphs.

Issue #511, opened 2022-08-31
canvas_item shader_type


novalis:
**Feature/enhancement description:**

For 2d textures in Godot, the `shader_type` is `canvas_item`, and output goes to `COLOR=` in the `vertex()` function (rather than `fragment()`).  There should be a new type of material, called something like "Dynamic 2D material" to support this.

novalis:
[material_2d.txt](https://github.com/RodZill4/material-maker/files/9472563/material_2d.txt)

Something like this -- but I have no idea what to do for Unity or Unreal, so ignore those parts.

RodZill4:
I guess this could just be an export option of the dynamic unlit material.

PR #518, opened 2022-09-15
Add godot 4 export templates

Arnklit:
This adds Godot 4 exports with heightmaps instead of depthmaps both for the Standard and ORM material versions.
I put the now 3 Godot export targets in a sub menu and the 2 Unity targets in a menu. Finally I saw that the SSS textures get exported, but not hooked up on the Godot material, so I hooked it up.
These were added for both "Static PBR Material" and "Static PBR Material Tesselated".

material_maker_r5mtveJOgj

EDIT: By the way, I made one more change. Currently the AO control from the material node is forwarded to ao_light_effect. This is incorrect, ao_light_effect, does not control the strength of AO, but how much AO shows up under direct light. I removed the line that sets this.

```ao_light_affect = $(param:ao)```


Calinou:
I'd use "Godot 3" instead of "G3" if we have the space.

Arnklit:
> I'd use "Godot 3/4" instead of "G3/4" if we have the space.

Yeah I did that first, the reason I changed it was that it will end up saying the entire string in the export window.

"Godot/Godot 3 Spatial" for example.

image

Alternatively we could also just remove the "G". What do you think @RodZill4 ?

RodZill4:
Yeah I think stripping what's before the / is the way to go.

BTW, I'll use that branch to improve other export targets:
- shader params requested by Nielsen
- dynamic unlit exported as canvas_item shaders (#511)

RodZill4:
Hmm I don't think I'll have time for shader parameters for 1.1... Will have to schedule this for 1.2. ;(

RodZill4:
Updated existing export targets and added new ones for Godot 4.
Added a prompt to overwrite material files when needed.

poiati:
> parameters

@RodZill4 have we added shader parameters already? Tks!

PR #540, opened 2022-10-19
New random algorithm for temporal anti-aliased view

paulofalcao:
Improved random algorithm for temporal anti-aliased view

Before
image

After
image


RodZill4:
Isn't there a better way of calculating dxy? Should be really simple (like 1.0/viewport_size?).

RodZill4:
Looks like the dxy and ratio variables are completely useless...

PR #544, opened 2022-10-26
Add Depth output to Normal Map node

myaaaaaaaaa:
Adds a "Depth" output whose scale is linked to the strength of the normal map, to streamline height-only workflows.

Intended usage:

The Normal Map node being proposed by this pull request:

### DIFF
--- a/addons/material_maker/nodes/normal_map2.mmg
+++ b/addons/material_maker/nodes/normal_map2.mmg
@@ -41,6 +41,24 @@
 			"from_port": 0,
 			"to": "switch",
 			"to_port": 1
+		},
+		{
+			"from": "math",
+			"from_port": 0,
+			"to": "gen_outputs",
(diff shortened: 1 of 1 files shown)


RodZill4:
It's a neat trick, but multiplying depth might be a bad idea for high values of Strength. If anything, I'd remove the multiplier node. Correlation between normal map and height map depends on the model anyway.

myaaaaaaaaa:
Thanks for the feedback! The idea is that the exact ratio between normal and depth can be adjusted with the material's `depth_scale` parameter, which can be set once based on the model's uv scale. Afterwards, the `Strength` parameter would be used to tweak heightmap scale with correct normals automatically applied, which I find especially useful for quick experimenting.

Workflows where normal and depth are less correlated would of course calculate depth from height themselves, and would not use this new `Depth` output.

PR #545, opened 2022-10-28
Add additional blending modes

paddy-exe:
+ Linear Light
+ Vivid Light
+ Pin Light
+ Hard Mix
+ Exclusion

The original functions are mostly only with vec3 return so I adapted the conversion from the existing shader functions to incorporate the opacity input

I recently added them to my addon library as well and I think MM could make  use of these as well. I took them from here:
https://www.shadertoy.com/view/XdS3RW


paddy-exe:
Hey, this would be my first contribution here. Just wanted to ask if technically I did everything correctly. I would add all the other blend modes from the source if everything is done right until now.

Arnklit:
Nice one @paddy-exe! I was actually needing hard mix today. You should probably add this into the documentation page as well. https://github.com/RodZill4/material-maker/blob/master/material_maker/doc/node_filter_blend.rst

Maybe we should consider adding more detailed information about each blend mode to that page or have a separate page with more info.

paddy-exe:
@Arnklit Thanks! Good to hear. I added the docs node additions. I think it would help if the blending modes would be explained directly in the blend node documentation even if it would mean a lot of information on one page. Perhaps an example picture showing the effect of the mode?

PR #547, opened 2022-11-04
Add a basic regression test that operates via checksums

myaaaaaaaaa:
Adds a shell script that tests for regressions in the given materials by exporting and then checksumming them. If even a single pixel changes after the script is rerun, the checksums will be different and you will be alerted.
Right now, the intended use case is manually testing for regressions when doing large invasive changes like mass refactors, and maybe basic sanity checking between version releases.

Also includes fixes to various problems with command line exporting that I encountered along the way.

Adding this into CI may be possible if software rasterization can guarantee consistent results across platforms, but is outside the scope of this PR.


myaaaaaaaaa:
Closing so I can rewrite this, sorry for the noise

PR #550, opened 2022-11-13
Html5 port

RodZill4:


start.gd
```diff
@@ -80,6 +80,28 @@ func _ready():
 	if loader != null: # check for errors
 		set_process(true)
 
+func start_ui():
+	if OS.get_name() == "HTML5":
+		var dialog = load("res://material_maker/windows/accept_dialog/accept_dialog.tscn").instance()
+		dialog.dialog_text = """
+			This HTML5 version of Material Maker has many limitations and is meant for evaluation only.
```
Zireael07, line 87: Nitpick: what are those "many limitations"?

RodZill4: No painting, no export, rendering 8 bits instead of 16 bits to name a few.
But this build is perfect for creating shaders (not textures).

RodZill4: Oh, and it's not some kind of "demo version" or something (MM is free anyway). Those limitations are due to the HTML5 platform.

Zireael07: I guessed :)

I think those three are the biggest and those could be listed in the dialog ("The HTML5 version doesn't support paintng, export or 16 bits rendering" is roughly as long as the current phrase")

PR #551, opened 2022-11-14
Remove internal copy of Godot binaries

myaaaaaaaaa:
Makes the repository significantly more lightweight (by ~4x!), although with the obvious caveat that I don't know what issues led to this in the first place and whether they're still around.


RodZill4:
Those are custom builds that make it possible to save .EXR image files. They are required.

PR #552, opened 2022-11-15
Fix texture memory leak in renderer.gd

myaaaaaaaaa:
One-line fix, opening a PR into #546 (`deps_manager`) to avoid merge conflicts in `master`.

### DIFF
--- a/addons/material_maker/engine/renderer.gd
+++ b/addons/material_maker/engine/renderer.gd
@@ -91,7 +91,7 @@ func render_material(object : Object, material : Material, render_size : int, wi
 	return self
 
 func render_shader(object : Object, shader : String, render_size : int, with_hdr : bool = true) -> Object:
-	var shader_material = $ColorRect.material
+	var shader_material = $ColorRect.material.duplicate(true)
 	shader_material.shader.code = shader
 	mm_deps.material_update_params(shader_material)
 	var status = render_material(object, shader_material, render_size, with_hdr)


addons/material_maker/engine/renderer.gd
```diff
@@ -91,7 +91,7 @@ func render_material(object : Object, material : Material, render_size : int, wi
 	return self
 
 func render_shader(object : Object, shader : String, render_size : int, with_hdr : bool = true) -> Object:
-	var shader_material = $ColorRect.material
+	var shader_material = $ColorRect.material.duplicate(true)
```
RodZill4, line 94: Hmm could you please explain this fix?

myaaaaaaaaa: Whenever a new set of textures is loaded - presumably every time a new material is opened - they're attached to `$ColorRect.material` in the function `render_shader()`. As `$ColorRect.material` is a long lived object (from what I can see, it stays in memory for the lifetime of the application), the textures will never be released until application exit. This is especially a problem as the new textures tend to have new names, meaning they won't replace old textures, and memory will continue to grow unbounded instead of simply staying at peak usage.

This is mostly a problem for large batch exports, although I assume that it would also happen to a lesser extent during regular editing as new textures (buffers?) are created and deleted.

The `duplicate()` fix makes it so that `shader_material` instead starts as a clean slate that can go out of scope once `render_shader()` exits, and therefore the textures attached to it can be cleaned up by reference counting.

RodZill4: The same problem could happen when calling render_material I guess.
Shouldn't we assign $ColorRect a new ShaderMaterial at the end of render_material instead?

myaaaaaaaaa: Is it ok to instantiate ShaderMaterials with `new()`? I assumed it was easier to just duplicate the original scene `$ColorRect.material`, since that's what `_ready()` was already doing. If that's the case, duplicating it in `render_material()` wouldn't work since it would just be duplicating a `$ColorRect.material` that already had textures attached in `render_shader()`.

From my testing (via monitoring RENDER_VIDEO_MEM_USED during batch exports), this was enough to fix it, and I wanted to avoid making more changes than necessary.

RodZill4: I modified the code (in the parent PR) to create new materials when needed. Thanks a lot for this fix.
Please don't hesitate to tell me if there are still leaks.

myaaaaaaaaa:
Great, thanks!

PR #555, opened 2022-11-17
Add highpass and sharpening functionality to fast blur shader

### LINKED ISSUE #527: Highpass filter request.
Highpass filter would be nice.

myaaaaaaaaa:
Implemented as two extra outputs to `fast_blur_shader.mmg`, since highpass is essentially the inverse of blurring and sharpening is highpass added back to the original image. Will attach the outputs to the user-facing node (`fast_blur.mmg`) along with documentation in a separate pr if this method is approved.

Fixes #527

From top to bottom:
Original image
Blurred output
Highpass output
Sharpened output
image


RodZill4:
This is a good idea.
Your screenshot shows the node renamed to High/Low pass, but this does not seem to be actually changed. I Think we should keep the old name, but add aliases for this node (high,low,pass) so it can easily be found.

RodZill4:
Oh, and BTW, please do the other tasks in this PR so it's easier for me to track changes.

myaaaaaaaaa:
Added to `fast_blur.mmg` now, with suggestions taken into account.

RodZill4:
Great, thanks.
Please don't be upset if I don't merge it soon, I generally keep node changes and new nodes for "just before release", so users who test dev builds don't add materials that use nodes-that-don't-exist-yet to the website (those materials wouldn't work with any release).
It's for 1.2 in any case.

myaaaaaaaaa:
No problem, thanks for the quick response!

Megalomaniak:
Would have really liked to see a full blown wavelets implementation but this is fine too, and perhaps something could still be achieved via mips?

Arnklit:
@myaaaaaaaaa  I know this is long since merged. I was just wondering what the reasoning was for centering the outputs on 0?

myaaaaaaaaa:
@Arnklit It simplifies the node trees for the use cases that I'm aware of. For example, color mixing looks like this:

PR #557, opened 2022-11-19
Adjust mmg format to make shader code more diffable

myaaaaaaaaa:
Right now, code blocks in shader nodes are bunched up into one giant string with \n literals when serialized. This PR proposes a new format where code blocks are instead split into arrays of strings, one per line, so that json formatting can make them more readable when viewed directly from github.

The proposed new format has been applied to beehive.mmg as a demonstration, see its diff for details. This also automatically adds a trailing newline for quality of life purposes, as mentioned in #51 and #119

Reading will be able to handle both the old and new formats, but writing will default to the old format unless the MM_WRITE_NEW_FORMAT environment variable is set. This way, the nodes inside the repo can be reformatted to get the benefits immediately, while backwards compatibility is kept between nodes and materials generated and shared by users. Presumably this can be removed in a later release once everyone has upgraded.

Includes a short gdscript that can bulk update .mmg files from the command line, eliminating the need to do it manually via gui.

### DIFF
--- a/material_maker/nodes/generic/generic.gd
+++ b/material_maker/nodes/generic/generic.gd
@@ -494,19 +494,46 @@ func save_generator() -> void:
 	while files is GDScriptFunctionState:
 		files = yield(files, "completed")
 	if files.size() > 0:
-		do_save_generator(files[0])
+		do_save_generator(files[0], generator)
 
-func do_save_generator(file_name : String) -> void:
+static func extract_verbose_values(path : String, data : Dictionary) -> Array:
+	var rv = []
+	var keys = data.keys().duplicate()
+	keys.sort()
+	for key in keys:
+		var v = data[key]
+		var mpath = path + "." + key
+		match typeof(v):
+			TYPE_STRING:
+				if "\n" in v:
+					data.erase(key)
+					rv.append("string:" + mpath.substr(1) + "\n" + v)
+			TYPE_DICTIONARY:
+				rv.append_array(extract_verbose_values(mpath, v))
+	return rv
+
+static func dict_tree_to_string(data : Dictionary) -> String:
+	data = data.duplicate(true)
+
+	var rv = ""
+	if OS.get_environment("MM_WRITE_NEW_FORMAT") != "":
+		var keyvals = extract_verbose_values("", data)
+		keyvals.push_front("")
+		rv = PoolStringArray(keyvals).join("\n########################################")
+
+	return JSON.print(data, "\t", true) + "\n" + rv
+
+static func do_save_generator(file_name : String, _generator : MMGenBase) -> void:
 	mm_globals.config.set_value("path", "template", file_name.get_base_dir())
 	var file = File.new()
 	if file.open(file_name, File.WRITE) == OK:
-		var data = generator.serialize()
+		var data = _generator.serialize()
 		data.name = file_name.get_file().get_basename()
 		data.node_position = { x=0, y=0 }
 		for k in [ "uids", "export_paths" ]:
 			if data.has(k):
 				data.erase(k)
-		file.store_string(JSON.print(data, "\t", true))
+		file.store_string(dict_tree_to_string(data))
 		file.close()
 		mm_loader.update_predefined_generators()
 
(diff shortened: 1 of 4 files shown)


RodZill4:
I do agree that JSON is a PITA (for git diffs for example), but I don't like this solution. There are other problems related to JSON (float precision for example), and I'm thinking about using a completely different format (or maybe 2 formats, one that is readable and a compressed format).

myaaaaaaaaa:
Fair enough, closing

RodZill4:
@myaaaaaaaaa You seem to have quite a few good ideas, why don't you join the discord server?
There's really a need for changing format (for several reasons).

myaaaaaaaaa:
Hello, thanks for the offer! I prefer to submit proposals alongside a minimal viable patch and then discuss them on github, and only follow up if they're approved. Especially since I tend to propose things that are easier to demonstrate than explain.

Also, sorry about the terse response. I closed this because, since a format change is already in the works, this PR (which is essentially its own format change) would just cause additional migration headaches for all contributors for no good reason.

RodZill4:
Thinking about this I maybe rejected this one too fast, it's not an elegant solution, but I'm not sure we have better formats/options than JSON, and it would solve the git diff mess when modifying a node.
I'll probably reopen it and change a few things (would like to move some of that code to loader.gd and not change the serialize/deserialize functions in nodes).

myaaaaaaaaa:
> (would like to move some of that code to loader.gd and not change the serialize/deserialize functions in nodes).

That makes more sense, actually. In that case, instead of changing the schema like I proposed, how about extending JSON itself with a multiline string type (encased with triple backticks instead of the normal quotes)? For now, it would be implemented as a regex preprocessor before JSON parsing in `loader.gd/load_gen()`, and a regex postprocessor after JSON printing in `generic.gd/do_save_generator()`.

Afterwards we can propose this extension as an optional argument in Godot's JSON functions.

myaaaaaaaaa:
@RodZill4 I've gone ahead and pushed the new version, can you check to see if this method is preferable?

RodZill4:
OK, had a look at the code and will test it this weekend.

RodZill4:
That new format is interesting and has advantages over JSON, but I prefer the first idea. It's not as elegant as I'd like, but it's the best we can have while staying compatible with JSON. I'll push a new version, please tell me if it's OK for you.

material_maker/nodes/generic/generic.gd
```diff
@@ -494,21 +494,27 @@ func save_generator() -> void:
 	while files is GDScriptFunctionState:
 		files = yield(files, "completed")
 	if files.size() > 0:
-		do_save_generator(files[0])
-
-func do_save_generator(file_name : String) -> void:
+		do_save_generator(files[0], generator)
+
+static func extract_verbose_values(path : String, data : Dictionary) -> Array:
```
myaaaaaaaaa, line 499: This can be moved to `loader.gd` now that `save_gen()` is there, although I think it's better to remove it along with `string_to_dict_tree_old()` unless there's a need to keep it?

scripts/fixup.gd
```diff
@@ -0,0 +1,21 @@
...
+	while file_name != "":
+		if file_name.ends_with('.mmg'):
+			print("converting "+file_name)
+			var file_path : String = "res://addons/material_maker/nodes".plus_file(file_name)
+			var file : File = File.new()
+			if file.open(file_path, File.READ) == OK:
+				var generator = loader.string_to_dict_tree(file.get_as_text())
+				file.close()
+				file.open(file_path, File.WRITE)
+				file.store_string(loader.dict_tree_to_string(generator)+"\n")
```
myaaaaaaaaa, line 19: It's better to call `loader.save_gen()` here since that has code to clean up the generator structure, and maybe also `loader.load_gen()` above so it's balanced 🙂

myaaaaaaaaa (commented):
ok, looks great! Just a few small things

scripts/fixup.gd
RodZill4: This is not as easy as it seems since actually loading a node relies on globals that are not available in editor mode.

RodZill4: Oh, never mind, it works.

PR #559, opened 2022-11-20
Add a "Colorspace Roundtrip" node for performing arbitrary computations in another colorspace

myaaaaaaaaa:
Converts the input to another colorspace (such as HSV), decomposes all channels, applies user-defined "subgraph parameters" to do arbitrary computations onto individual channels in that colorspace, recombines them, and converts it back to RGB for the final output.

These "subgraph parameters" are supplied in the same way as Iterate Buffer's "loop subgraph". Marked as draft since this is a throwaway proof-of-concept implementation, the real one will require modifications to the dependency checking rules, which I will open a PR for if this concept is approved.

All these names are, of course, temporary and open to suggestions.

My head hurts.

Example usage: Remove shadows by blurring only the V of HSV
example

Crude wiring schematic of this node:
schematic


Arnklit:
I don't think having nodes that you plug back into themselves makes sense outside iteration nodes. I think this is a confusing workflow and it's worth making people add one more node to avoid it, so I would suggest just adding "RGB to HSV" and "HSV to RGB" nodes. Which definitely seems like a good idea.

RodZill4:
I do agree with @Arnklit , it would be easier if this feature is a pair of RGBtoX/XtoRGB nodes. Still a very cool (and necessary) feature. :)

myaaaaaaaaa:
Since there's demand for a simpler interface for HSV, I'll redesign the Color Conversion node too. I need it anyway for a subgraph-based reimplementation of this "Colorspace Roundtrip" node, since I'm not smart enough to get input->output dependency tracking working through the variables in the current glsl implementation.

@Arnklit Would putting this node under a new "[Lambda/](https://en.wikipedia.org/wiki/Anonymous_function)" category (along with "Iterate Buffer", currently under Miscellaneous/), implied to be for advanced users who are willing to work with these "lambda subgraphs", be enough to address your concerns?

@RodZill4 Here's a mock-up of the new Color Conversion node, does this look okay?

colconv

Arnklit:
That looks great @myaaaaaaaaa.

Is there anything special you can do with the roundtrip idea that you can't do with just using two conversion nodes? Because I might just be missing something?

It would be up to Rod, if certain nodes in misc should be split off in another category, but it might be overkill with a new category, it could just be a sub category in misc.

RodZill4:
I think 2 nodes (from RGB/to RGB) would be easier.
I don't think this needs a new category or something, those nodes can be in "Filters".

myaaaaaaaaa:
Finished, but depends on #564 to enable self-connections

PR #567, opened 2022-12-01
Add "Log Scale" parameter to Scale node

myaaaaaaaaa:
Makes it easier to set scale to extreme values, independently of aspect ratio.

[log_scale.webm](https://user-images.githubusercontent.com/103326468/204961965-94c23efa-82a4-42b5-b3aa-86d5153ce6c4.webm)


RodZill4:
I think this should be a new node, with a single scale parameter. Note that holding Control and Alt make it possible to tune parameters more precisely.

PR #568, opened 2022-12-02
Correctly emit signal when render queue becomes empty

myaaaaaaaaa:
Was causing command line exports to hang, likely affects normal exporting too.


RodZill4:
Dammit, clicked wrong button!

PR #569, opened 2022-12-03
Fix memory leak in dependency tracker

myaaaaaaaaa:
`delete_buffer()` was deleting `buffer_name` from `buffers`, but not `dependencies_values`


RodZill4:
Ooooh, that was a pretty bad one.
Did you actually notice leaks, or audit the code? I was counting on dependency cleanup line 77, but if this didn't work, there may still be another problem that causes useless dependency data to remain.

myaaaaaaaaa:
8GB of RAM makes leaks more noticeable 🥲

PR #571, opened 2022-12-08
Sets default uv size for cylinder to 3 on x axis

Disar:
Makes the uv mapping less stretched on the cylinder by default


Calinou:
Are you sure this is the correct value? The texture still seems stretched, judging from the appearance of CylinderMesh in the Godot editor.

### UV Scale (3, 2)

image

Compare this with `Vector2(2, 4)`:

### UV Scale (2, 4)

image

PR #574, opened 2023-01-01
Various new nodes, with a focus on workflow nodes

obsidianical:
This adds various new nodes, among them a lot of new workflow-related nodes, fixes to the Output node, two grunges and a filter.


Cole-Allen:
Any chance we could split this into smaller chunks for ease of review?

obsidianical:
I've changed my opinions on workflow nodes (among other reasons because of the separation of normal and height maps, and some of the really cursed blending stuff), so I probably should close this. Also, those nodes are probably mostly significantly outdated as well. 

I'm working on a bunch of other stuff for MM whenever I have time, but the current state of workflow nodes is at best only usable for stylized game dev stuff, since they make it *really* hard to make high detail and realistic textures.

PR #588, opened 2023-02-23
Add spherize node

williamchange:
This adds a new variadic transform node which distort input image(s) as if they were wrapped around a sphere:

spherize_node_demo

Example outputs:

sample_outputs


williamchange:
- Remove scale parameter
- Rename Offset > Center

update

PR #593, opened 2023-03-23
Random Weave Node

Arnklit:
This adds an alternative weave node that has randomization controls.

Godot_v3 5 2-stable_win64_iGoBHQZnv6


RodZill4:
- Shouldn't the "Random" parameter be named Randomness?
- Sometimes, a thread can be all-top or all-bottom. Maybe an option to avoid this?

PR #594, opened 2023-04-01
Add $rndi expression function

Arnklit:
This adds `$rndi` as alternative to the `$rnd` expression function.

The `$rndi` returns whole numbers by rounding the random function.

This is really practical when using expressions in fields that expects whole numbers.

https://user-images.githubusercontent.com/4955051/229279532-c6f1c7f2-bcb7-45ae-ba8f-a6ff8ec915ba.mp4

### DIFF
--- a/addons/material_maker/common.shader
+++ b/addons/material_maker/common.shader
@@ -42,3 +42,7 @@ vec3 hsv2rgb(vec3 c) {
 float param_rnd(float minimum, float maximum, float seed) {
 	return minimum+(maximum-minimum)*rand(vec2(seed));
 }
+
+float param_rndi(float minimum, float maximum, float seed) {
+	return round(param_rnd(minimum, maximum, seed));
+}
--- a/addons/material_maker/common.shader
+++ b/addons/material_maker/common.shader
@@ -44,5 +44,5 @@ float param_rnd(float minimum, float maximum, float seed) {
 }
 
 float param_rndi(float minimum, float maximum, float seed) {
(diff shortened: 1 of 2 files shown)


Arnklit:
@RodZill4 one thing I was not sure about was this magic value. My understanding was that it was how many chars would be added, so because my function name was one char longer I should use 18. 

image

RodZill4:
> @RodZill4 one thing I was not sure about was this magic value. My understanding was that it was how many chars would be added, so because my function name was one char longer I should use 18.
 The value is actually not important, it's just there to make sure 2 $rnd calls in the same node don't roll the same value. You could have reused rnd_offset instead.
 Note that I need to have a look at that code since random seeds changed a little bit.

Arnklit:
> The value is actually not important, it's just there to make sure 2 $rnd calls in the same node don't roll the same value. You could have reused rnd_offset instead.
Note that I need to have a look at that code since random seeds changed a little bit.

ah grand, yeah maybe you could look it over, and clean it up a bit, I wasn't super sure of all the moving parts, so I might not have done it the most elegant way.

Arnklit:
Actually just thinking about this, it might be wrong to use "round" for this.

If we do round(rand(min, max)) 

Isn't there less change that min and max get picked than the values in between? I'll do a test of this and see if I'm right, I think it does need to be `floor(rand(min, max+1.0))` to have the correct spread.

Arnklit:
Yeah it was incorrect to use round. I went over to floor, I also went over to using the same offset as rnd was using, rather than having a separate one.

PR #602, opened 2023-04-16
Remember Preview3D custom mesh across restarts

myaaaaaaaaa:
Makes it easier to work on textures that are tied to specific geometry.


RodZill4:
Hmm too bad this does not remember default meshes and their parameters. I'll keep this for 1.4

PR #604, opened 2023-04-21
Fix regression in Image node's aspect ratio

myaaaaaaaaa:


RodZill4:
Damn I broke this in such a stupid manner...

PR #642, opened 2023-11-12
Make sure graph edit background doesn't darken preview

Jowan-Spooner:
This adds a "StyleBoxEmpty" override to the graph edit node. Previously the preview was slightly darkened due to the background.


Jowan-Spooner:
I will close this. The Graph preview has other problems now, but this is way to outdated ;)

PR #650, opened 2023-12-17
Make SDF boolean nodes variadic

williamchange:
Turns several SDF boolean nodes variadic so they can work with multiple shapes.

2d_boolean_variadic

2d_smoothboolean_variadic


williamchange:
3d sdf boolean variadic nodes

3d_boolean_variadic

3d_smoothboolean_variadic

williamchange:
This also swaps the inputs around so the first input is the background instead of the foreground (addresses a [suggestion](https://discord.com/channels/784201835334074389/789948291517382759/1098930067352068116) from @Tar0x on Discord) as it can be confusing when working with multiple inputs(variadic nodes).

Before:

2d_before
3d_before


After:

2d_after
3d_after

williamchange:

PR #656, opened 2024-01-25
Fix broken Japanese glyphs

williamchange:
Fixes broken(i.e. disconnected segments) glyphs in the Japanese Glyphs node:

Hiragana: あ(a), す(su)
Katakana: イ(i), ウ(u), ク(ku), ケ(ke), タ(ta), チ(chi), オ(o)

fix


RodZill4:
Hmm is this related to 1.4 not rendering exactly like 1.3 or did you just improve those glyphs?

williamchange:
> Hmm is this related to 1.4 not rendering exactly like 1.3 or did you just improve those glyphs?

Nothing wrong with the rendering, just improving the glyphs

PR #661, opened 2024-02-09
Add Classic Kuwahara filter node

williamchange:
Classic version of the Kuwahara filter, adapted from https://code.google.com/archive/p/gpuakf/

node_sample

Sample Images

sample

samples_gif


williamchange:

PR #667, opened 2024-02-27
Improve spherize node

williamchange:
sph_demo

Some improvements for the Spherize node, mainly:
- Add center x/y control widgets
- Don't clamp input image values
- Prevent taking square root with negative values(this now takes the absolute before sqrt), which caused some visual issue on Mac OS - I previously made the node on 12.6.x which works fine but it worked differently in 14.3


williamchange:

PR #660, opened 2024-02-07
Add Anisotropic Kuwahara filter node

williamchange:
[Anisotropic Kuwahara filter](https://docs.blender.org/manual/en/4.0/compositing/types/filter/kuwahara.html) based on Blender's Kuwahara compositor node ([compositor_kuwahara_anisotropic.glsl](https://projects.blender.org/blender/blender/src/branch/main/source/blender/compositor/realtime_compositor/shaders/compositor_kuwahara_anisotropic.glsl))

node

**Sample image**

sample

PR #668, opened 2024-02-27
Add Generalized Kuwahara node

williamchange:
Adds Generalized Kuwahara filter node, based on [GarrettGunnell's implementation](https://github.com/GarrettGunnell/Post-Processing/blob/main/Assets/Kuwahara%20Filter/GeneralizedKuwahara.shader). This filter performs better than [Classic Kuwahara](https://github.com/RodZill4/material-maker/pull/661) but is still inferior to [Anisotropic Kuwahara](https://github.com/RodZill4/material-maker/pull/660) at preserving directional image features.

node_img

image_sample

Comparison with other Kuwahara filters

kuwahara_vs


williamchange:

PR #670, opened 2024-03-04
Fix missing short/long description for some nodes

williamchange:
I've noticed some nodes have missing descriptions.

Most of these are group nodes and they don't have a proper name so the internal name is used for the tooltip instead (e.g. fast_blur, morphology2, sd_mask_to_sdf). This PR adds the missing short and long descriptions (based on the documentation) for them.

Before

before

After

after


williamchange:
Added short/long descriptions:
- Normal Blend
- Normal to Height
- Convert From/To RGB
- Smooth MinMax

PR #673, opened 2024-03-21
Enhanced Flexible Tab UI

Takanu:
Changes
- Increased padding for the content container to be more legible and visible.
- Added padding between the buttons to prevent misclicks and improve visual clarity.

Reasons?
- Improved legibility of tab text.
- Attempts to unify the tab design with those used in the graph menu (haven't found the code for that yet to validate👌 )

Before:
image

After:
image


Takanu:
Almost certainly, im not familiar with the codebase and its been a bit since I've used Godot.  Ill get to it when im able to.

Jowan-Spooner:
I think this has been covered by my layout/theme PR and can be closed. We can further adjust the actual values, but it is now implemented and these values can be changed from the theme resource.

Takanu:
I think so too, i've made the changes to the theme files but I think it's best included as a separate PR.

PR #696, opened 2024-06-05
Fix generalized kuwahara node

williamchange:
Fixes Generalized Kuwahara node (was outputting nothing - white on Master)


williamchange:
Checking this again on a recent build (428966662728c410d2034c98f1923d41d693509f) now outputs black, but this should still fix the issue:

https://github.com/user-attachments/assets/348175bc-199e-43f7-915d-a63ee461d843

PR #701, opened 2024-08-14
Fix seams on tiled textures when using certain nodes

BlueCube3310:
Changes the texture samplers' repeat properties to repeat by default to make them function the same way they did in previous versions. This fixes the seams that appear when using most processing nodes:

| 1.3 | 1.4a | 1.4 PR |
|----|------|--------|
|||1_4PR|


RodZill4:
Hi! Can you please provide the test case? there's still a diffrerence between the 1.3 and 1.4 screenshots and I'd also like to understand this (plus see if there are more required changes for this problem).

BlueCube3310:
A very simple test case: [repeat_test.zip](https://github.com/user-attachments/files/16647734/repeat_test.zip)

The file consists of a checkerboard node connected to HBAO. The output looks like this on different versions (most noticeable with a plane preview mesh):

1.4 alpha:
test

1.3:
test2

BlueCube3310:
This can also be seen on many existing materials, such as https://www.materialmaker.org/material?id=998

A noticeable line appears in 1.4 alpha:

1.3:
k3

1.4:
k4

This may also fix #722

PR #708, opened 2024-09-04
Experimental Scaling Approach

Jowan-Spooner:
This PR tries to solve multiple issues:
- Scaling is messy in sub-windows or separate windows
- Higher Scale have blurry icons
- Icons are fixed colored, meaning they can't adapt to the theme

It solves these by not using the Window.content_scale_factor and instead scales the theme values up and reloads theme icons from svg when the scale changes. During this loading certain icon colors can be changed as well.

## EnhancedTheme Resource
This whole logic is implemented by the EnhancedTheme resource. It can be used in place of a theme and takes a base_theme which can then be modified with a scale and color_swap settings.

## Requirements & Drawbacks
This system only works with scalable icons (svg). This meant changing the way many icons are implemented (WIP). This system also ONLY scales up the theme. Any relevant constants outside the theme (like minimum sizes, magic numbers in drawing etc.) are NOT automatically updated. This means more scripts need to listen to _notification(NOTIFICATION_THEME_CHANGED) and use mm_globals.ui_scale to update these things. 

It also means all icons should be defined in the theme instead of being set directly on the nodes. I've added a helper script to make this easier on buttons. Basically this boils down to just using `get_theme_icon("IconName", "MM_Icons")` instead of load("res://...icons.tres")

I've started these necessary changes, but there are A LOT of buttons with icons in places that I don't know very well so this is definitely still a WIP (especially because I also need to recreate many icons and would like to find a more consistent style at the same time).

This was a bunch of work, especially because inkscape and godot didn't like to collaborate, there is some bug in ThorVG... 

If anyone would like to test this, especially on a HiDPI monitor, that would be very cool!


Calinou:
Note that Godot 4.x will eventually get a built-in solution for this: https://github.com/godotengine/godot/pull/86022

Jowan-Spooner:
Yeah, I saw this. I'm a bit fearful of "eventually". But this system is design in a way that should make it pretty easy to take it out again with not too much work, if this change eventually makes it to godot (and actually fully solves this problem). The biggest work so far seems to be replacing all the icons with svgs, a change that is necessary anyways.

Also while that PR might solve the scaling issue, it wouldn't allow us to change icon colors, which this PR would. From previous discussion with the MM people, it seems that a more customizable theme would be appreciated and having this kind of theme pre-processor in place would enable that pretty easily as well.

That said, I'll wait for Rod to take a look and see whether he like this or not :)

PR #710, opened 2024-10-22
Fix other themes

Jowan-Spooner:
- Removes most of the old themes
- Renames "Modern" to "Default Dark"
- Recreates the "Classic" theme
- Adds a simple theme-variant system
   - Creates a "Default Light" theme variant of the Default Dark theme

ToDo:
- follow up Icons PR, where all icons are replaced with a single svg icon sheet.
- Some more cleanup in the themes folder
- Possibly implement theme-scaling as an alternative to content_scale scaling.


Calinou:
> Possibly implement theme-scaling as an alternative to content_scale scaling.

Out of curiosity, why is that needed? https://github.com/godotengine/godot/pull/86022 should make a custom approach unnecessary once it's merged, as it'll be possible to easily use hiDPI-friendly icons out of the box.

Jowan-Spooner:
@Calinou Yeah, I know, I would love to use that. But it's been stale for another half year. I'll try to keep my workarounds easily undoable because of course the built-in solution would be best, but it's just not there yet. So I will create workarounds until then. Rod has considered removing the multi-window support because of how buggy scaling is with sub-windows and I think that would be a real bummer, so I've been experimenting with alternative ways to get better scaling. Anyways the theme-scaling isn't in this PR anyways, it's just something that could be possible with this approach.

Also this has other benefits, such as recoloring icons, which isn't possible otherwise.

PR #717, opened 2024-11-06
Fix dialogs for saving/loading export targets

myaaaaaaaaa:
`window_title` was changed to `title`


myaaaaaaaaa:
Fixed in https://github.com/RodZill4/material-maker/commit/ba4b020c9823a4139231b954b317988ea8d4b812

Issue #479, opened 2022-07-06
Add .glTF format as an export option


RobotRoss:
**Feature/enhancement description:**
Add the ability to export completed texture to glTF 2.0 compliant .glTF file

.glTF is a next-generation interchange format spearheaded by Khronos
[site](https://github.com/KhronosGroup/glTF)

Godot also has support for exporting scenes to .glTF
[site](https://godotengine.org/article/introducing-the-godot-gltf-2-0-scene-exporter)

Adding the .glTF format would enable Material Maker to be used in applications that are outside the scope of the existing export options.

For now, I'd like the ability to export just the completed textures to glTF, but in future it'd be useful to also export any models which are present in the scene to the same glTF file.

clairem-sl:
Now that Godot 4 is out, will this be developed?

Calinou:
> Now that Godot 4 is out, will this be developed?

[Material Maker is being ported to Godot 4](https://github.com/RodZill4/material-maker/tree/godot4), but the goal is to get existing features working first.

RobotRoss:
> is glTF supposed to be used as a standard for sharing materials?

There's no reason why it couldn't be - Some platforms accept .glTF for materials, but not MaterialX - hence why libraries such as [https://github.com/kwokcb/glTF_MaterialX](glTF-MaterialX) exist.

That said, having MaterialX as an export option isn't a bad idea, but it doesn't replace glTF as an export target.

ColonelThirtyTwo:
In particular, Blender can import glTF files natively but does not seem to be able to import MaterialX. Same with the Bevy game engine.

PR #718, opened 2024-11-06
Implement a basic glTF exporter for materials

myaaaaaaaaa:
Adds a `GLTF/Plane` export target alongside the standard ones like `Unity/HDRP`, `Godot/Godot 4 Standard`, etc. The export includes a scene with a simple plane mesh for easily previewing the material.

* Related to #479


myaaaaaaaaa:
Friendly ping

PR #719, opened 2024-11-07
Adjust height of preview cylinder

myaaaaaaaaa:
Changes the cylinder height to `PI / 2`, which eliminates the UV stretching seen in the current model.


RodZill4:
I will replace the cyclinder with a chamfered one (and make sure height is correct).

myaaaaaaaaa:
In that case, I think it would be better to output unstretched UVs so that the height can remain adjustable:

RodZill4:
I just fixed rounded cylinder so it does not have stretched UV. Thanks!

PR #721, opened 2024-11-08
Add a variadic color picker node

### LINKED ISSUE #531: Select specific color node.
One of the more useful thins would be to have "select color" node that similar in tones range fasion would allow to target specific color from source to be used as a mask with with/contrast of course.

myaaaaaaaaa:
Useful for ID map workflows:

For more examples, see https://www.google.com/search?q=id+map+substance&udm=2

Resolves #531

### DIFF
--- a/addons/material_maker/nodes/color_pick.mmg
+++ b/addons/material_maker/nodes/color_pick.mmg
@@ -1,0 +1,77 @@
+{
+	"name": "color_pick",
+	"node_position": {
+		"x": 0,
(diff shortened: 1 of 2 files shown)


RodZill4:
Ooooh, this one is a good idea!
Not sure about the name, though, would probably deserve the "mask" word in its name.

myaaaaaaaaa:
How about "Colors to Masks"?

Naming is hard.

PR #755, opened 2025-02-12
Add improved normal2height node

Qendolin:
Add [my normal2height node](https://www.materialmaker.org/material?id=1059) to the base library.
The default node works fine for very simple geometric shapes, but for real normal maps it produces very strong streaking artifacts.
This improved node is based on Materialize's height from normal generator.

Probably the sampling sequence and performance can still be improved. But it's not bad. Also I haven't done any spell checking. 
I'll keep this as a draft until initial feedback. 

| Normals | Simple | Improved | Reference |
|---|---|---|---|
| normal_to_height_compare_normal_map_convert_2_map | normal_to_height_compare__3_old | normal_to_height_compare__5_new | N/A |
| normal_to_height_compare_normal_map_convert_2_roof_normal | normal_to_height_compare__4_roof_old | normal_to_height_compare__5_roof_new | normal_to_height_compare__7_roof_ref |
| normal_to_height_compare_normal_map_convert_2_window_normal | normal_to_height_compare__3_window_old | normal_to_height_compare__5_window_new | normal_to_height_compare__7_window_ref |


Qendolin:
I'm not really sure if it makes sense to keep the existing, simple node

Qendolin:
I should note that the strength and base height parameters don't work exactly like in the existing node. It's just a simple `value * strength + base`

Qendolin:
One issue I've found is that high sample counts (4k or more) often result in crashes with textures above 512 resolution. However this is more of an issue with MM than this node in particular. If Iterate Buffers exposed the iteration number, I could use them to spread the calculation across multiple invocations and the crash could likely be avoided.

Qendolin:
I have a version in the works that uses an iterate buffer. That makes it more stable. And I've improved the quality further.
Future Work (not in this PR): See how awesome bump does it.

lycorys-emporium:
Tested this locally, merges in clean and works well on the latest commit of the main branch (edf1a141f944fc39747c47bd582df3347739c372) and on Godot 4.5.1.

Only issue I found was, after building the project and adding the node to an existing material to create a snow effect, it didn't generate _anything_ usable, generating what appeared to be a pink-and-black checkerboard pattern, something I'm assuming is a placeholder if something goes wrong?

I didn't get a screenshot of it, however, and the issue went away on saving and reopening the material. I could've messed up a build step somewhere as well since I was manually doing what the Action does for the dev build (couldn't find a build script, there doesn't seem to be one).

After that, the material worked fine and the heightmap was usable and worked as expected, converting my material's normal from a cloud noise node into something usable compared to the broken, streaky simple node that already exists.

Merging this PR in would really help! The original streaking artifacts issue still happens in the existing simple normal to height node in the latest release as of this comment, version 1.4, so having this would make it possible to generate height maps for anything the simple one wasn't good for.

Issue #727, opened 2024-11-28
"Create Library" window is broken


NotArme:
**Material Maker version:**
Issues found both on oficial 1.4a and latest release from master

**OS/device including version:**
Windows 11

**Issue description:**
- Window that pops-up when creating a library cannot be closed with the usual **X** button
- Prompt for library selection is labeled `Library file`, but the button is `click to open a path`, making it not clear whether i'm supposed to choose a file or path (I think it should be a path)
- When clicking the button, the file select screen opens up on fullscreen, and it keeps flashing black (could not capture that issue on the video)
- The file select screens asks for a json file instead of a path, so it's impossible to create a new library without manually creating an empty .json file first

- This last one isn't a bug, but it would be nice to be able to input the file creation path as a string without having to use the file select button, example from blender: 
![image](https://github.com/user-attachments/assets/13238b8e-a850-45c8-b38a-0fbcaf307107)

**Steps to reproduce:**

https://github.com/user-attachments/assets/869003b3-72fa-4d55-ab85-64d4c7f436b4

PR #759, opened 2025-02-23
Overhaul custom library creation window

NotArme:
This PR attempts to fix all bugs outlined on [Issue 727](https://github.com/RodZill4/material-maker/issues/727).

## Changes
An editable line was added to make it easier to manually add a path or change the filename
When only a directory is chosen for the path, a file name will now be automatically created based on the chosen library name
"Ok" button is now disabled if library name is empty
The file dialog will now correctly ask for a path instead of an existing file
Library creation window layout and size was slightly modified to accomodate changes

X button will now correctly close the window
Added window title "Create Library"

image
image

### DIFF
--- a/material_maker/panels/library/create_lib_dialog.gd
+++ b/material_maker/panels/library/create_lib_dialog.gd
@@ -68,3 +68,4 @@ func enter_info(value : String = "") -> Dictionary:
 
 func _on_VBoxContainer_minimum_size_changed():
 	min_size = $MarginContainer.get_combined_minimum_size()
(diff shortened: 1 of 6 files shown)


RodZill4:
Not sure it should be resizable... But it's less use friendly if it's not. :(
image

NotArme:
I think it's definitely a plus to have it being resizable horizontally at least, I'm not sure if you can easily do that on godot while keeping vertical resizing locked, I'll take a look at it

PR #763, opened 2025-03-02
Improve macos disk image background

williamchange:
Improve labels legibility in disk image for the mac os build and fix background position/scaling

Before / After:
disk_image_preview


Calinou (approved):
That's an interesting solution to the problem, but if we can't change the text color ourselves, it's the best we can do.

williamchange:
> change the text color

Probably not possible. I think most dmg's I came across either use a light image background or no background at all (which then the labels would be black/white depending on the system's light/dark theme setting)

PR #767, opened 2025-03-05
Normalize bevel range for japanese glyphs node

williamchange:
Normalize bevel range for Japanese Glyphs node so it's actually a factor (i.e. 0.0-1.0) based on the radius

Before:

before

After:

after


RodZill4:
Since it changes the behavior of the node, it should be a new mmg file. ;)

Issue #761, opened 2025-03-01
Popup text appears much smaller than other UI text (on Mac)


dgholz:
This might be related to #723, but looks like a different issue to me.

**Material Maker version:**
1.4a3, from [`bc702ff`](https://github.com/RodZill4/material-maker/commit/bc702ff4d46ce886a04580ef72120909c9dfcb24)

**OS/device including version:**
MacBook Air M1, MacOS Sonoma 15.3.1

**Issue description:**
The text in popup windows, like preferences, is much smaller than other text.

<img width="837" alt="Image" src="https://github.com/user-attachments/assets/c1bead1f-d61c-475b-886e-ecddf8d4b01a" />

<img width="311" alt="Image" src="https://github.com/user-attachments/assets/5314436e-ed32-4cfd-94b0-36360f4b9fd7" />

<img width="339" alt="Image" src="https://github.com/user-attachments/assets/b7682c8a-f044-4cf8-8da0-70eb5edcf415" />

**Steps to reproduce:**
Go to Edit menu, click Preferences. Or in the Preview3D panel (or set background to 3D), go to the Model menu & click Configure.

dgholz:
The exit popup looks fine!

<img width="339" alt="Image" src="https://github.com/user-attachments/assets/db626f69-8bcb-40c4-987c-eeb179d6e577" />

Calinou:
This is likely a case of `Window.content_scale_factor` not being applied to Window instances created by Material Maker for its popups.

LastPlaceProdiG:
It's not just mac, I've got the same problem on windows.

williamchange:
All popups(incl. Undockable panels) sizes should already be fixed for hidpi displays in rc2 (since  https://github.com/RodZill4/material-maker/commit/cb3012c78599d36ce6d13032e14d8607905bd866)

dgholz:
can confirm, text size in popups matches the rest of the UI text in v1.4rc4

PR #780, opened 2025-03-17
Fix scales for most dialog/pop-ups for hi-dpi displays

williamchange:
Fix dialog/pop-up scales for hi-dpi displays, for now this addresses the popups mentioned in https://github.com/RodZill4/material-maker/issues/761 and the curve edit pop-up. I tested this on Macbook Pro M1 (Sequoia 15.3.2):

**Preferences:**
image

**Curve editor:**
image

**Mesh config popup:**
image


williamchange:
By now this should cover most dialog/popups, except the undockable panels

PR #813, opened 2025-04-13
Add per-point/node gizmo size control

williamchange:
Adds an additional control point (P4.xy) and per-node/control gizmo size customization (only P1-P4 for now). This adds another field in the data model - `control_size` to store the size of the gizmo

https://github.com/user-attachments/assets/06c582b2-0381-4a13-88fa-e361696eea2c


RodZill4:
Hmm can you think of a way we could make it possible to choose the big/small icon per point control and per node?

williamchange:
> Hmm can you think of a way we could make it possible to choose the big/small icon per point control and per node?

I guess it could be an additional dropdown next to 'control' from the node editor

williamchange:
https://github.com/user-attachments/assets/c6e73a40-2eea-4eaf-80b8-cb59ed4725ce

PR #814, opened 2025-04-13
Fix sdf3dcolor shader error by setting default input value to 1.0

williamchange:
Likely a mac(m1)-specific issue
glitch


williamchange:
moved to #815 for a slightly better fix

PR #829, opened 2025-04-19
Preview 2D alpha checkers customization

williamchange:
alphacheck


Calinou:
Do we really need full color customization on the checkerboard, or can we just expose a few brightness options (dark/medium/bright) like GIMP does?

williamchange:
> Do we really need full color customization on the checkerboard, or can we just expose a few brightness options (dark/medium/bright) like GIMP does?

That's fair, I thought color could be useful as it might be possible some users will make a pattern that's close to what the checkers look like making it harder to see, but perhaps brightness/checker size presets could already be good enough

williamchange:
checker_v2

Issue #263, opened 2021-01-29
Way to quickly reset entry fields to default (Right click?).


some9000:
Right click on entry fields does nothing, so, perhaps, it could reset the value to default (or 0)? Would be great.

Tar0x:
Sure, I think right clicking it should have dropdown menu for more options in the future :)

PR #843, opened 2025-05-01
Reset/Expression editor via float edit context menu

williamchange:
Resolves https://github.com/RodZill4/material-maker/issues/263

This also moves the expression editor into the context menu

float_edit_context


williamchange:
Added copy/pasting

https://github.com/user-attachments/assets/2cd5a3ff-0d4f-448f-8a26-34bb06a35ed2

PR #867, opened 2025-05-08
Remove white border for doc node images using imagemagick

williamchange:
Improve doc page appearance for nodes by getting rid of the white border from the node screenshots using ImageMagick to turn it transparent:

```shell
magick $in \
    \( +clone -gravity center -shave 10x10 -bordercolor none -border 10x10 \) \
    \( -clone 0 -fuzz 10% -transparent white \) -delete 0 -composite $out
```

This isn't perfect as it expects the ports to not be white so it does not work well on reroute and the switch node, but could probably cover all nodes if another background color is used


williamchange:
I guess it might probably the best to just use the background color from the doc pages when doing the screenshots

PR #870, opened 2025-05-10
Fix splines export

williamchange:
New font resource `Rubik-Bold.tres` was added in https://github.com/RodZill4/material-maker/pull/841 but not exported


williamchange:
Resolved with 950590c5f71707e6a56a39f6156be83ab1c97054

PR #882, opened 2025-05-12
Fix library section button click

williamchange:
Before / After:

https://github.com/user-attachments/assets/2709c38a-57f8-4bd6-977e-ed34af8f9278

### DIFF
--- a/material_maker/panels/library/library.gd
+++ b/material_maker/panels/library/library.gd
@@ -234,10 +234,7 @@ func _on_Section_Button_pressed(category : String) -> void:
 		if item.get_text(0) == category:
 			item.select(0)
 			item.collapsed = false
-			for node in tree.get_children(true):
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Out of curiosity, why to the center? I actually remember making it go to the top on purpose (though it seems it broke at some point), because it felt much more useful to have it scroll to the top.

williamchange:
> Out of curiosity, why to the center? I actually remember making it go to the top on purpose (though it seems it broke at some point), because it felt much more useful to have it scroll to the top.

@Jowan-Spooner for reference v1.3 looked like this for me so I tried something that's at least better than that and that it's a simple enough fix (but that's fair, I will push a fix)

https://github.com/user-attachments/assets/37e1b511-d4fd-4c7d-8987-9ea01511cf13

PR #889, opened 2025-05-14
Make splashscreen mm url link clickable

williamchange:
URL has a 'Pointing Hand' CursorShape (hinting that it is clickable), but does nothing on click


williamchange:
Resolved with https://github.com/RodZill4/material-maker/commit/8f4b191acc1cbe749fc061a319072394e90e844c

PR #903, opened 2025-05-22
Linux: Option to disable MMB paste to comment box

williamchange:
https://github.com/user-attachments/assets/5334ddef-9701-4482-83c5-632410648a8b


Calinou:
Should this only affect the comment text box, or all text boxes within nodess in Material Maker (e.g. shader expresssion)?

PR #937, opened 2025-06-08
Fix links offset changing to/from classic theme

williamchange:
Port positions are changed but links do not connect to them, this tries to force an update

Before / After:

https://github.com/user-attachments/assets/ebabad89-527c-406b-b043-262269c6481d


williamchange:
Closing since this is no longer an issue

Issue #643, opened 2023-11-13
Use Native OS File Dialog


werdnahull:
**Feature/enhancement description:**
Godot's built-in file open/save dialogs are not great. It would be nice to use the native OS dialogs for open, save, and export, which would greatly increase workflow. Something like [https://github.com/98teg/NativeDialogs](https://github.com/98teg/NativeDialogs) would likely work.

Calinou:
Godot 4.2 and later has built-in support for native file dialogs. Material Maker is being ported to Godot 4, so it'll surely be using 4.2 or later.

RobotRoss:
+1 to this - The current file dialog works if your computer has a single hard drive, but if you have resources stored across drives or on network shares, it's downright awful.

PR #956, opened 2025-06-13
Add preference to use system native file dialogs

williamchange:
- Resolves https://github.com/RodZill4/material-maker/issues/643

https://github.com/user-attachments/assets/742f3702-ea79-4e15-9c51-156def73cab5

https://github.com/user-attachments/assets/129d2fcb-3d8e-47ba-b0f5-88294585d687

### DIFF
--- a/material_maker/globals.gd
+++ b/material_maker/globals.gd
@@ -52,6 +52,7 @@ const DEFAULT_CONFIG : Dictionary = {
 	auto_size_comment = true,
 	graph_line_curvature = 0.5,
 	graph_line_style = 1,
+	ui_use_native_file_dialogs = true,
 }
 
 
--- a/material_maker/windows/file_dialog/file_dialog.gd
+++ b/material_maker/windows/file_dialog/file_dialog.gd
@@ -5,7 +5,7 @@ var _content_scale_factor: float = 1.0
 var left_panel = null
 var volume_option = null
 
-const DIALOG_HACK : bool = true
+var dialog_hack : bool = true
 
 signal return_paths(path_list)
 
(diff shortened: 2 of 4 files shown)


williamchange:
On Linux (Fedora 42) it's a bit buggy, probably just enabling it isn't enough

https://github.com/user-attachments/assets/d46d6598-1601-46ea-a771-4d94d43a6de4

jkcsg:
What is the current status of this?

On Windows, the native file dialog is much easier to handle that what Material Maker 1.4 ships with. The current one does not have proper Icons and even though it has some sort of Favorite section, this is quite underdeveloped and not synchronized with my Favorite links in Windows Explorer. I understand that on some platforms the build in Dialog may be a useful addition, but on Windows it is strictly worse.

It seems to be already successfully implemented, but I cannot see this feature in the official release. What is stopping this to be integrated?

material_maker/globals.gd
```diff
@@ -30,6 +30,7 @@ const DEFAULT_CONFIG : Dictionary = {
 	auto_size_comment = true,
 	graph_line_curvature = 0.5,
 	graph_line_style = 1,
+	ui_use_native_file_dialogs = false,
```
Calinou: I suggest enabling this by default, as it's what most people expect for GUI applications these days.

williamchange:
Addressed review comment

> What is the current status of this?

I'm mostly waiting for @RodZill4 to merge the following since it's pending in the milestone and it will result in a conflict if this get merged first (most preferences-related PRs will, so I am marking them as drafts for now), but I guess if this is more desirable i can always fix it afterwards

- https://github.com/RodZill4/material-maker/pull/969

williamchange:
This could probably be a blocker (mac)
- https://github.com/godotengine/godot/issues/113685

PR #974, opened 2025-06-21
Fix reroute size not updating when changing themes

williamchange:
Current / PR

https://github.com/user-attachments/assets/ca9847b0-7915-4e85-b19f-4ddb09142bfd


williamchange:
Superseded by https://github.com/RodZill4/material-maker/pull/975

PR #988, opened 2025-06-29
Round library section buttons

williamchange:
rounding


Jowan-Spooner:
Note that I had added new icons for the sections on the new icon sheet and had planned to replace these with actual buttons with icons instead of these texture buttons for maximum stylability (and making it easier to potentially recolor the sections if we ever wanted to do so), but never got to it. 
Feel free to do that if you like the idea or not if you don't :) Your efforts on the UI side are awesome and appreciated! Hopefully I'll find the time some day to return and be of help.

williamchange:
Tried adding them(as buttons with the new icons), I think most of them are bearable, except noise which is definitely very hard to read with their current resolution

image

williamchange:
Updated library colors based on @Jowan-Spooner's UI Proposal

[Palette(via coolors.co)](https://coolors.co/903047-7e7e32-238277-4e4e94-a64b9d-477a47-246283-94633c)

vs

PR #998, opened 2025-07-03
Show/hide scrollbars based on ui activity

williamchange:
Show scrollbars only when they are needed for the control, e.g. panning in GraphEdit, scrolling, zoom in/out etc.

https://github.com/user-attachments/assets/4dd0e676-c986-4e45-8f47-9c94959b3aa9


Calinou:
From an UX perspective, I'm personally not fond of autohiding scrollbars. I would rather have the scrollbar decrease in width (so you still know where you are in the page/graph), but not disappear entirely.

RodZill4:
I agree this is a rather big change with very little benefit.

PR #1064, opened 2025-07-31
Fixed environment preview thumbnail generation

williamchange:
Restores v1.3 behavior by adding a surface material override with same roughness/metallic settings


williamchange:
- Merged into #1063

PR #1078, opened 2025-08-07
Remove unnecessary `@tool` in `_view/_editor` scripts

williamchange:


williamchange:
Closing for now as per a note from Rod:
> ...at some point we might want a link between MM and other tools (including Godot) and we might need those controls in tool mode.

PR #1112, opened 2025-08-30
Link node to unlit material output

williamchange:
Adds a shortcut <kbd>Shift</kbd>+<kbd>Alt</kbd>+<kbd>LMB</kbd> to link nodes to the unlit material node

Linking(performing the shortcut on) the material restores the previous material/connections

https://github.com/user-attachments/assets/7bfb2343-9b37-42a2-aeda-ccdcb9fdbc8d

### DIFF
--- a/material_maker/panels/graph_edit/graph_edit.gd
+++ b/material_maker/panels/graph_edit/graph_edit.gd
@@ -24,6 +24,9 @@ var need_save_crash_recovery : bool = false
 var top_generator = null
 var generator = null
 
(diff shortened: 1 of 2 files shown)


williamchange:
> ...it should be a 3d preview feature that does not modify the graph.
This feature would be very useful when texturing models, to show masks and baked maps.

As per a note from Rod the preferred solution is to implement this as a feature in the 3d preview. Will be keeping this up as a reference for the time being

PR #1134, opened 2025-10-08
Fix double file dialog for linux (i.e. flatpak)

williamchange:
This should not have any effect on other platforms, but on Linux (i.e. flatpak) this fixes the native filedialog being shown twice.

Opening this PR as per Calinou's suggestion via 
- https://github.com/flathub/io.github.RodZill4.Material-Maker/pull/17


williamchange:
Closing this as it's included in 
- https://github.com/RodZill4/material-maker/pull/956

PR #1175, opened 2025-11-05
Fix VRAM leak in Preview2DPanel

DevFika:
**Issue**
When previewing a buffer node in the 2D preview panel, the texture wasn’t getting released from VRAM after deleting the node.
The old material shader kept a reference to the texture in one of its parameters, so it never got freed.

**Fix**
Added a check in Preview2DPanel.set_generator() to clean up the old texture reference before switching shaders.
If the old generator buffer texture is still set on the material, it now sets that shader parameter to null, letting the VRAM free up.

**Related**
There’s also a case where deleting a buffer node after deactivating its preview slot leaves behind an orphaned Preview2D node, which prevents VRAM cleanup since that orphaned node keeps a reference to the old buffer texture.
#917

**Before:**
vram_buffer_before_pr

**After:**
vram_buffer_after_pr

### DIFF
--- a/material_maker/panels/preview_2d/preview_2d_panel.gd
+++ b/material_maker/panels/preview_2d/preview_2d_panel.gd
@@ -110,6 +110,14 @@ func get_shader_custom_functions():
 
 
 func set_generator(g : MMGenBase, o : int = 0, force : bool = false) -> void:
+	if generator and is_instance_valid(generator):
+		if not is_instance_valid(g) or g != generator:
+			var old_id = generator.get_instance_id()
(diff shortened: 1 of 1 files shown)


RodZill4:
I am not sure this fix will solve the problem in all cases. What happens if you select a node that's connected to the buffer (and can be worse with several buffers) and delete both nodes ?

BTW, are you on Discord?

DevFika:
Ah, you're right. this doesn’t handle connected nodes keeping those references, I overlooked that. A better fix is probably needed.

Yes, I'll send you a message!

Issue #1174, opened 2025-11-05
Add support for the ARM platform


PuKYC:
**Feature/enhancement description:**

Release official builds for Linux ARM and Win ARM

PR #1176, opened 2025-11-05
Enable arm64 builds for linux

williamchange:
- Partially implements https://github.com/RodZill4/material-maker/issues/1174

Adds Linux arm64 CI build step

Edit: Still WIP, mostly missing steam export and need to fix export presets to add arm64 export for linux


Calinou:
Note that Windows ARM really needs to be using Direct3D 12 to work well, as I understand it. There's no native Vulkan support on that platform, so it would have to go through a translation layer, which adds a performance overhead and can introduce bugs.

williamchange:
I think this pr would just be Linux for now, I don't really have a Windows ARM machine to properly test a build

PR #1191, opened 2025-11-14
Made godot export target names consistent

williamchange:
Makes it consistent across both static pbr/displacement targets

`G3` > `Godot 3`
`G4` > `Godot 4`


RodZill4:
Although I would like to homogeneise target names even more, this is a step in the right direction.

PR #1216, opened 2025-12-18
Added node inputs swapping (Alt+S)

williamchange:
Add node input swapping similar to Blender(Alt-S), based on suggestion via discord

> allow swapping node connections by holding shift+drag or something similar (in blender its alt). sometimes when using the Blend node i want to swap background & layer1 and then you have to move one then grab the output of the other node and reconnect that one.

https://github.com/user-attachments/assets/439ac6e8-9f1a-4c71-beba-3ad5584ab421


ppmpreetham:
This is super helpful!

ppmpreetham:
It's merged let's goooooo

PR #1217, opened 2025-12-19
Exposed number as an input map for 7seg/roman no.

williamchange:


RodZill4:
This must be a new version of the node since it could break existing materials (the 1st input becomes the 2nd etc.).

williamchange:
> This must be a new version of the node since it could break existing materials (the 1st input becomes the 2nd etc.).

Done

PR #1218, opened 2025-12-20
Adding parameter labels to most nodes that were missing labels

DeerUwU:
(hope this PR is acceptable)
to address some labeling inconsistencies, i've added labels to a lot of nodes that were previously missing labels. 

if this PR gets accepted, i would like to create a follow-up PR to reorganize parameters and inputs to make sure that node inputs are no longer placed next to unrelated parameters (similar to how blender does it.). this should simplify working with nodes and prevent potential confusion.

here are some of the changed nodes:
image


RodZill4:
Adding missing labels is acceptable.
Reordering parameters is a problem for users who know MM. Might be acceptable in some cases.
Reordering ports will break existing materials, so this is not acceptable.

In your screenshot:
- shape and sphere are OK`
- Profile would need a deeper rework IMHO
- Smooth boolean is worse with your change because of variadic aspect
- Mix Maps is OK, even though the change does not help a lot (the node mixes maps, so yeah, the inputs are maps)

And that kind of PR should come with updated node screenshots (they are generated automatically).

williamchange:
You might want to update the node screenshots again without the dots background

Issue #1219, opened 2025-12-22
star shape has artifacts when set to 4, 12, 20 or 28 sides


DeerUwU:
**Material Maker version:**
issue occurs in 1.4 and 1.5pre (tested in build from 19/12/2025)

**OS/device including version:**
Windows 10

**Issue description:**
star shape node creates unwanted pixels when setting side count to specific values
<img width="1360" height="998" alt="Image" src="https://github.com/user-attachments/assets/e3a5c83e-1c42-411a-992d-72ae538020af" />

**Steps to reproduce:**
Create Shape Node, set to "star" and set side count to 4, 12, 20 or 28.

williamchange:
What GPU are you using?

DeerUwU:
nvidia rtx 3070ti, tested on game-ready driver version 576.52 & 591.59. ive done some more testing and noticed that the issue is not visible in the 2d preview if you set the view filter to "none" instead of a specific resolution. it does however always appear in the 3d preview and exported images

PR #1221, opened 2025-12-23
Fixed star shape artifacts

williamchange:
- Should resolve https://github.com/RodZill4/material-maker/issues/1219

[fix.webm](https://github.com/user-attachments/assets/b2ab5f5e-b124-43a0-82d2-315c92469901)


RodZill4:
Not sure why this fix works. :/

PR #1234, opened 2026-01-02
Updated license one last time

williamchange:
Uses `-present` in place of the current year.


RodZill4:
I'm not sure I like this change, it's low effort to update this once per year.

Issue #1154, opened 2025-10-21
Linux build freezes/crashes when opening external/floating tabs or tooltips


Kujyssisi:
**Material Maker version:**  
v1.4

**OS/device including version:**  
Arch Linux (EndeavourOS)

**Issue description:**  
On some Linux systems, external tabs or windows that are not part of the main/root tab can cause the program to freeze or crash. In Godot, the usual fix for this is using single-window mode, but there is no such option in the Material Maker preferences.  

The only reliable workaround I’ve found is either using the Windows build via Wine or avoiding opening any popups/tooltips in the Linux build.  

**Steps to reproduce:**  
It seems to happen randomly; I haven’t figured out a consistent trigger. However, it is not just happening to me and other Linux users have reported similar issues with external/floating tabs on godot.

hypotensor:
Seems to have just happened to me from a tooltip popup. On EndeavourOS/Arch using Wayland.

CharlyAlbin:
Happening to me as well when displaying a submenu from the TopBar Menu in the MainWindow.

**Material Maker version:**
v1.4

**OS/device (including version):**
Arch Linux (CachyOS Wayland)

**Steps to reproduce:**
1. Launch Material Maker.
2. Click on File (top left).
3. Repeatedly hover over Load recent and Export material.

CharlyAlbin:
Same problem in the Godot editor.

https://github.com/godotengine/godot/issues/103249

It seems to be an issue with Vulkan. Launching the editor with `--rendering-driver opengl3` fixes the crashes. Does MM have a similar option?

williamchange:
> Same problem in the Godot editor.
> 
> https://github.com/godotengine/godot/issues/103249
> 
> It seems to be an issue with Vulkan. Launching the editor with `--rendering-driver opengl3` fixes the crashes. Does MM have a similar option?

MM 1.4 onwards is Vulkan-only

Xury46:
Same issue on Fedora 43 KDE Plasma Desktop on Wayland.

kpfriesen:
> Same problem in the Godot editor.
> 
> [godotengine/godot#103249](https://github.com/godotengine/godot/issues/103249)


A reply here said their issue was fixed after moving to the nvidia 590 driver. I was consistently having this issue but I've also been unable to reproduce this crash since upgrading to 590.

EndeavourOS/3070 TI/Wayland

BitCrack:
> > Same problem in the Godot editor.
[...]

PR #1306, opened 2026-02-13
Support single-window mode

williamchange:
Adds support for single-window mode

This may be useful for Linux users where multi-window mode can be unstable(at least until godot resolves the problem),

- i.e. https://github.com/RodZill4/material-maker/issues/1154
- https://github.com/RodZill4/material-maker/issues/1385

This PR
- Respects `--single-window` argument via command line
- Adds option in preferences to enable single-window mode (defaults to disabled)
- Adds theme styles(embedded border) for single-window mode
- Fixes for window/popup scale/positions under single-window mode
- Remove unnecessary context menu popup position fixes

### DIFF
--- a/addons/flexible_layout/flexible_layout.gd
+++ b/addons/flexible_layout/flexible_layout.gd
@@ -542,11 +542,14 @@ class FlexWindow:
 	var overlay : Control
 	
 	func _init(main_control : Control, first_panel : Control = null):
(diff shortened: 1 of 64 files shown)


Funkschy:
This would be amazing to have. On wayland the application frequently freezes permanently when overlays/tooltips are on opened

Issue #240, opened 2020-11-27
Proposal - Add a Gabor Noise Texture Node.


14AUDDIN:
Add the Gabor Noise Texture type, this type of noise can be used to create fabrics and add surface imperfections:
https://www.youtube.com/watch?v=f3_EZbrUUQs
https://graphics.cs.kuleuven.be/publications/LD11FSGN/

Calinou:
Are there open source implementations? The reference implementation has a restrictive license:


> ### Commercial Use
>
> The code is licensed under a non-commercial license (see LICENSE.txt). Permissions beyond the scope of this license are be available from the authors.

14AUDDIN:
There's this patch for Blender although it is written for OSL instead of GLSL:
https://developer.blender.org/D3495
![image](https://user-images.githubusercontent.com/42714433/100552589-27363a00-3280-11eb-95aa-4bb200433e1b.png)
![image](https://user-images.githubusercontent.com/42714433/100552595-2c938480-3280-11eb-9b88-bc77df08f38c.png)

some9000:
I am a super-noob at this fantastic app, but even to me it felt really easy to create something close to your example pictures using the built-in tools. Changing the values/merging a couple of these should produce basically the exact same thing.

![Cloth_A](https://user-images.githubusercontent.com/17021558/105815835-2b4eb180-5fbc-11eb-90f2-ded45d3b7306.png)

![Cloth_B](https://user-images.githubusercontent.com/17021558/105815846-2db10b80-5fbc-11eb-8654-c5fd8abe97a4.png)

Sure, it's not as clean as just having a single neat texture node, but you get more control over the final result... (Cannot attach a source file, but here's a screenshot of how I made it:

![material_maker_jO3qsw96Cp](https://user-images.githubusercontent.com/17021558/105816246-bb8cf680-5fbc-11eb-9ce5-2ef3061e4fed.png)

Calinou:
> (Cannot attach a source file, but here's a screenshot of how I made it:

You can compress the source file into a ZIP archive and upload it here :slightly_smiling_face:

some9000:
> You can compress the source file into a ZIP archive and upload it here 🙂

Alright, I'll keep that in mind for the next one. This makes good immersion into the methods, though :)

RodZill4:
If you find a nice combination of nodes you think is reusable, you might want:
- to group them into a single node (select nodes you want to group and hit Control-G).
[...]

PR #1321, opened 2026-03-04
Add gabor noise node

williamchange:
https://github.com/RodZill4/material-maker/issues/240

image


williamchange:
Noise doesn't really tile in certain cases, probably needed more work - closing this for now

vs

PR #1343, opened 2026-03-20
Alternative inputs for graph panning/camera orbiting

williamchange:
context via discord
> I can't rotate the object in the 3D panel. I think it is bound to a middle mouse button, which I do not have.

This adds `Shift + LMB` for graph panning and `Alt + LMB` for 3d orbit

### DIFF
--- a/material_maker/panels/graph_edit/graph_edit.gd
+++ b/material_maker/panels/graph_edit/graph_edit.gd
@@ -335,6 +335,9 @@ func _gui_input(event) -> void:
 			accept_event()
 			lasso_points.append(get_local_mouse_position())
 			queue_redraw()
+		elif (event.button_mask & MOUSE_BUTTON_MASK_LEFT) != 0 and event.shift_pressed:
+			scroll_offset -= event.relative
+			accept_event()
 
 
 func get_padded_node_rect(graph_node:GraphNode) -> Rect2:
--- a/material_maker/windows/environment_editor/camera_controller.gd
+++ b/material_maker/windows/environment_editor/camera_controller.gd
@@ -20,7 +20,8 @@ func _ready():
 
 func process_event(event : InputEvent, viewport : Viewport = null) -> bool:
 	if event is InputEventMouseMotion:
-		if event.button_mask & MOUSE_BUTTON_MASK_MIDDLE != 0:
+		if (event.button_mask & MOUSE_BUTTON_MASK_MIDDLE != 0
+			or (event.alt_pressed and (event.button_mask & MOUSE_BUTTON_MASK_LEFT) != 0)):
 			if event.shift_pressed:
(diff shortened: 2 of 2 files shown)


jamez9:
Is it possible to also add zoom equivalents for the graph editor and 3d preview? Trying to use with a laptop and trackpad, I don't think there is a way to zoom these with hotkeys and no mouse. Presumably the same issue is had from a mouse user without a zoom wheel.

Even just being able to use the standard + and - keys would be super useful to be be able to zoom in or out of any view.

williamchange:
In the [3d preview](https://github.com/RodZill4/material-maker/blob/0fe99355c49ec2dd21d3ad1626d7a10161dd7f37/material_maker/windows/environment_editor/camera_controller.gd#L61) you should be able to zoom with the trackpad pinch gesture, I think in the graph view the same gesture also zooms, which OS are you using?

williamchange:
I tested on mac the gesture works

jamez9:
Thanks, I was able to get pinch zoomi working on my Win10 laptop but only for the 2d and 3d previews, not the graph editor. My Linux Mint laptop was not able to do this, but I don't think it has a trackpad capable of doing zoom. 

The graph editor does have the + and - zoom buttons at least, but not sure if there is any way of binding these to hotkeys? No zoom widget exists in the 2d or 3d preview windows though. Would be good to unify the zooming ablities on all 3 panels at some point in the future.

williamchange:
- There's a PR for zoom shortcuts for the graph editor https://github.com/RodZill4/material-maker/pull/929

williamchange:
>  not the graph editor.

Does scrolling on the trackpad work?

jamez9:
On my win10 laptop in the graph view , when using the pinch zoom, it does seem to scroll up and down in an odd way, but in general it seems confused. I guess my issue with relying on pinch to zoom in general, is that it is very specific to the hardware. While it is standard on Mac (which have very nice trackpads), older and cheaper Windows laptops don't all support it. Personally, I am not a fan of gestures on trackpads and prefer hotkeys.

Great to see the PR for zoom htokey support. Is it possible this could also be added to the other views? Btw, Godot itself does have crtl + and - for zooming in it's 2d views.

williamchange:
For now you can use ctrl + lmb drag to zoom the 2d views

williamchange:
- Opened PR: https://github.com/RodZill4/material-maker/pull/1350

williamchange:
> The graph editor does have the + and - zoom buttons at least, but not sure if there is any way of binding these to hotkeys?

It seems these buttons are bound to the same key which is a bug, I opened a PR to fix them so they will be bound to Ctrl+/-, though I think those only works if they are visible(you can pin the 'panels' by right clicking on them)

image


- https://github.com/RodZill4/material-maker/pull/1349

PR #1395, opened 2026-04-18
Fix stack overflow when editing input portal link

williamchange:
Fixed a case where stack overflow can occur(in editor) when i.e. editing the highlighted input portal to "aperture_1"

This only happens if the node is connected by an output portal whose source the edited link will reference from

image


williamchange:
- resolved in https://github.com/RodZill4/material-maker/pull/1414

PR #1407, opened 2026-04-28
Add interface fonts options

### LINKED ISSUE #1406: Improved Interface Font
I am running Windows 11 25H2.
Material Maker version: 1.6
Monitor Resolution: 2560x1440 (27inch)
Display Scaling: 100%

The current interface font does not look right. I have attached an image along side of file explorer in windows 11, the fonts are too big and does not use system font. I tried changing the UI scale to reduce the size, but then the font is not easy to read.

Image

**Feature/enhancement description:**
Similar to UI scale, it would be nice to have an interface font option to pick an default font and size for the interface. By default it should use system font(at least thats what most apps use). And probably need some minor spacing adjustment.

williamchange:
- Resolves https://github.com/RodZill4/material-maker/issues/1406

Adds custom font/size options

image

.
If system fonts are enabled, [`sans-serif`](https://docs.godotengine.org/en/stable/tutorials/ui/gui_using_fonts.html#system-fonts) will be used be in place of the custom main/medium/bold fonts that are set in preferences.

These settings could probably be applied without restarting, but I am getting a lot of `"fd" is null` errors if I do that so this is what I have for now.


Calinou:
> If system fonts are enabled, [sans-serif](https://docs.godotengine.org/en/stable/tutorials/ui/gui_using_fonts.html#system-fonts) will be used be in place of the custom main/medium/bold fonts that are set in preferences.

It might look better to use a tailored set of system font names according to the platform, as per this proposal:

- https://github.com/godotengine/godot-proposals/issues/5767

This can be implemented in scripting by checking for the current OS and using the relevant font name.

williamchange:
SF Pro wasn't available on my machine so I have to get it from https://developer.apple.com/fonts/ to test on macOS

It seems to have problems with MSDF rendering so I only added Windows changes for now

image

williamchange:
Added option to switch between msdf/oversampling

image

Calinou:
I think exposing a MSDF toggle makes sense to fix fonts with self-intersecting outlines, but exposing oversampling is kind of a double-edged sword. Setting it too high will make small text look grainy (only zoomed-in graphs will benefit from oversampling).

Fixing rendering of MSDF fonts with self-intersecting outlines requires https://github.com/godotengine/godot/pull/61555. It works, but it adds a lot of binary size for niche functionality, so we'd prefer avoiding it if possible.

williamchange:
Removed oversampling options

image

PR #1482, opened 2026-07-09
Fix extra spacing in gradient color picker popup

williamchange:


williamchange:
Combined into https://github.com/RodZill4/material-maker/pull/1481

PR #1497, opened 2026-07-13
Update node images and optimized images for documentation

williamchange:
Update node images and optimized images via `oxipng -o6 --strip safe --zopfli`


williamchange:
the savings is probably not worth the large change

PR #1553, opened 2026-09-04
Enable _process only when needed for post-process menu

williamchange:


williamchange:
- Superseded by https://github.com/RodZill4/material-maker/pull/1556

PR #1560, opened 2026-09-12
Properly set audio driver to "Dummy"

williamchange:


Calinou (approved):
Looks good to me. The setting path was likely changed in Godot 4.

Issue #1215, opened 2025-12-17
Slovenian translation


anderlli0053:
Hello!

I'd like to add Slovenian/Slovene localization, i see that this is supposed to be the directory for translations ➡️ https://github.com/RodZill4/material-maker/tree/master/material_maker/locale/translations but where are the source strings?

also what format is the main one .PO or .CSV?

After the translation i can open pull request, since you don't use the Crowdin or any other similar platform to do so.

williamchange:
Hi, both .po and .csv can be loaded into MM, I think the ones in the repo arent up to date but you can generate one via running [this script](https://github.com/RodZill4/material-maker/blob/master/material_maker/locale/generate_po_template.gd) in Godot which will give you a template for po:

[material-maker-all.pot.zip](https://github.com/user-attachments/files/24226833/material-maker-all.pot.zip)

anderlli0053:
> Hi, both .po and .csv can be loaded into MM, I think the ones in the repo arent up to date but you can generate one via running [this script](https://github.com/RodZill4/material-maker/blob/master/material_maker/locale/generate_po_template.gd) in Godot which will give you a template for po:
> 
> [material-maker-all.pot.zip](https://github.com/user-attachments/files/24226833/material-maker-all.pot.zip)

So i used the shell script with the same name ( _generate_po_template.sh_ ), but since i am not that familiar with the GNU's formats like this POT/PO/MO, i am not sure, does the biggest one contain more strings since it is way bigger and which one to translate then? The generated smaller .POT i assume?

williamchange:
pot is the template to start from,  I will recommend using po edit than editing the files manually https://poedit.net/ once done you export it as a .po

I haven't used the shell scripts myself but the larger one should have all strings

anderlli0053:
> pot is the template to start from, I will recommend using po edit than editing the files manually https://poedit.net/ once done you export it as a .po
> 
> I haven't used the shell scripts myself but the larger one should have all strings

Yes, i figured it has something to do with the quantity of the strings in each template, so basically, [the larger one](https://github.com/RodZill4/material-maker/blob/master/material_maker/locale/material-maker-all.pot) is the **go-to** for translating?

I know it sounds dumb, I'd rather ask than do something that ends up being pointless or wrong.

williamchange:
> > pot is the template to start from, I will recommend using po edit than editing the files manually https://poedit.net/ once done you export it as a .po
[...]

PR #1563, opened 2026-09-13
Add Italian translation

albanobattistella:


williamchange:
Hi, for translations you will need to host this file somewhere with a link (ideally a github repository here) then contact @RodZill4 so they can update the registry where MM will pull translations from
- see this comment for details https://github.com/RodZill4/material-maker/issues/1215#issuecomment-4244628809

PR #1565, opened 2026-09-15
Android Support

williamchange:
- Similar to https://github.com/RodZill4/material-maker/pull/1561 but for Android

### CI
- ~~Adds a `debug.keystore` in the repo only for singing debug builds,~~ for release exports secrets are used (i.e. `KEY_ALIAS` and `KEYSTORE_PASSWORD`), the release keystore is expected to be in Base64 under `KEYSTORE`

Following only applies when running on Android

### Splashscreen
- Disabled animated backgrounds due to performance concerns
### Flexible Layout
- Disabled undocking as it doesn't make much sense on a small screen, but maybe this could be made optional via preferences
### Preferences
- Adds an option to keep screen on
- Disabled single window mode option

### General
- Long press: right click
- Double tap: double click
- Back button: close dialogs, exit app when none shown
### Graph
- Two finger tap: undo
- Two finger pan: graph offset
- Two finger pinch: zoom
- Three finger tap: redo
### 2D previews / Reference
- One finger pan: view offset
- Two finger pinch: view scale
### 3D previews
- One finger pan: orbit
- Two finger pinch: zoom

## Ideas/Todos
- on screen modifier buttons for actions like noodle knife tool, lasso selection(similar to Godot editor)
- on screen customizable floating shortcut bar to perform menu actions (similar to GodSVG)
- tip text are mostly for desktop so they might not make sense on a touch device, though they could probably be replaced/mapped easily(long press: right click, double tap: double click) but this need some thinking
- ~~status bar controls could be moved to the menu bar(top right) to maximize screen space~~, notifications like "saved to ..." etc. could be displayed as toast messages
- the default layout probably have to be redefined for mobile
- deal with file path uri when using native file dialogs, mostly on displaying them
- a documentation page for common gestures, similar to the shortcut page but for touch inputs


Calinou:
I'm curious how well it works when using the Mobile renderer, given the lower rendering precision. (You can get Mobile to have the same precision as Forward+ by enabling **HDR 2D** on viewports, but this affects UI rendering due to different alpha handling.)

> Adds a `debug.keystore` in the repo only for singing debug builds, for release exports secrets are used (i.e. `KEY_ALIAS` and `KEYSTORE_PASSWORD`), the release keystore is expected to be in Base64 under `KEYSTORE`

Godot should be able to automatically generate the debug keystore now. Is that not being done when exporting from the command line? If so, we should probably address this on the engine side.

It's not a big deal either way since you've also configured release exports, but for debug-only export workflows, we could perhaps streamline this a little.

williamchange:
testing locally on mac Looks like it's signed, I based the ci from GodSVG so I didn't check
```
Signer #1:

Certificate #1:
Owner: CN=Godot, OU=Godot Engine, O=Stichting Godot, C=NL
Issuer: CN=Godot, OU=Godot Engine, O=Stichting Godot, C=NL
```

williamchange:
v0

PR #1566, opened 2026-09-16
Allow node field edits with a single click.

BarronKane:
Float fields now explicitly enter edit mode on focus while preserving text selection. Users can type numeric values after a single click instead of needing multiple rapid clicks.

Currently when you click a node field, it highlights the text as if you can begin typing in it, but you cannot. 

A second click will make the field typable, but gets rid of the highlight so you then have to worry about cursor position.

A third click in rapid succession will then highlight the text again and make it typable. 

This fix highlights the selection AND makes it typeable simultaneously. A second click then drops into cursor position un-highlighted for precise positioning. This aligns the expected behavior with most node-graph editors in the industry today.


williamchange:
which OS are you on? 
this already works for me

https://github.com/user-attachments/assets/de71289d-eaef-45b4-aea2-cd03c476b62c

BarronKane:
> which OS are you on? this already works for me
> 
>  test.mp4

@williamchange 

Oh this is interesting. I'm on linux with wayland kde.

BarronKane:
@williamchange 

What's interesting to me is that the original click handler switches the field to "editing", calls grab_focus(), and immediately selects text-- but it never *tells* godot's LineEdit to enter its separate internal editing state. That's what I added. Idiomatically, this fix would still be the semantically correct option.

